PR #3, opened 2016-01-11
Add errors list support to UnsatisfiedInputsException

Horneth:


src/main/scala/wdl4s/UnsatisfiedInputsException.scala
```diff
@@ -3,3 +3,6 @@ package wdl4s
 class UnsatisfiedInputsException(message: String, cause: Throwable) extends RuntimeException(message, cause) {
   def this(message: String) = this(message, null)
 }
+class UnsatisfiedInputsExceptionWithErrors(message: String, cause: Throwable, val errors: Seq[String] = Seq.empty) extends UnsatisfiedInputsException(message, cause) {
```
geoffjentry: Since you say it's with Errors then errors should be a NEL, there should be no possibility of it being empty

geoffjentry:
What is the motivation for this?

Horneth:
So currently all errors are smashed up into a big `errors.mkString("\n")` which becomes the message of the exception.
As part of the Cromwell ticket to provide clean json error responses in the API I thought it would have been nice to be able to have something like
```
 "status": "fail",
 "message": "Workflow e4c98006-4831-42e5-b98e-3a90110d24fb failed to process inputs",
  "errors": [
    "Required workflow input 'updatesymobls.H.r' not specified.",
    "Required workflow input 'updatesymobls.C.j' not specified.",
    "Required workflow input 'updatesymobls.H.q' not specified."
   ]
```
Doing this cleanly is currently not possible, the only way would be to `exception.getMessage.split("\n")` which is ugly and still doesn't fully give you only the validation errors.
We can also decide not to do this and send back the raw message but it's not very elegant IMO:
```
"status": "fail",
 "message": "Workflow e4c98006-4831-42e5-b98e-3a90110d24fb failed to process inputs\n\nRequired workflow input 'updatesymobls.H.r' not specified.\nRequired workflow input 'updatesymobls.C.j' not specified.\nRequired workflow input 'updatesymobls.H.q' not specified."
```

geoffjentry:
@Horneth One of the things that caused me to raise an eyebrow was that you're not actually using the `cause` anywhere in wdl4s.

Horneth:
@cjllanwarne was also chosen by the weel

src/main/scala/wdl4s/UnsatisfiedInputsException.scala
```diff
@@ -1,5 +1,18 @@
...
+  override def getMessage = {
+    s"""$message\n${errors.list.mkString("\n")}"""
+  }
+}
+
 class UnsatisfiedInputsException(message: String, cause: Throwable) extends RuntimeException(message, cause) {
   def this(message: String) = this(message, null)
 }
+
+class RuntimeValidationException(val message: String, val errors: NonEmptyList[String]) extends RuntimeException(message) with ThrowableWithErrors
```
geoffjentry: I thought you had a use case where there was just the message? If not, disregard this :)

geoffjentry:
@Horneth I think I misunderstood your explanation last night and was picturing this as being the errors coming from RuntimeAttributes, but you're talking about errors for coercions and declarations - apologies. Looking at this more closely I think what I'd recommend is simply replacing the UnsatisfiedInputsException with your new thing. The two places it's used do not use the `cause` and they both always have errors (I think ...)

src/main/scala/wdl4s/UnsatisfiedInputsException.scala
Horneth: Yes those cases use the pre-existing `UnsatisfiedInputsException`, which is actually correct because they're thrown when a input is missing or cannot be coerced.

geoffjentry: Right. See my other comment, it turns out I misunderstood what you were doing. I (think I) understand now

Horneth: Yes basically `UnsatisfiedInputsException`s are thrown each time an input fails to validate.
We then get a list of those exceptions that get mapped to a `RuntimeValidationException`, with the `errors` being the messages from the `UnsatisfiedInputsException`s

src/main/scala/wdl4s/WdlExceptions.scala
```diff
@@ -0,0 +1,15 @@
...
+trait ThrowableWithErrors extends Throwable {
+  def message: String
+  def errors: NonEmptyList[String]
+
+  override def getMessage = {
+    s"""$message\n${errors.list.mkString("\n")}"""
+  }
+}
+
+class UnsatisfiedInputsException(message: String) extends RuntimeException(message)
```
cjllanwarne: I think this should be singular - UnsatisfiedInputException

cjllanwarne: Any reason why this extends RuntimeException?

src/main/scala/wdl4s/WdlNamespace.scala
```diff
@@ -81,8 +82,9 @@ case class NamespaceWithWorkflow(importedAs: Option[String],
         optionValue = tryValue.get if tryValue.get.isDefined
       } yield key -> optionValue.get)
     } else {
-      val message = failures.values.collect { case f: Failure[_] => f.exception.getMessage }.mkString("\n")
-      Failure(new UnsatisfiedInputsException(s"The following errors occurred while processing your inputs:\n\n$message"))
+      val errors = failures.values.collect { case f: Failure[_] => f.exception.getMessage }
```
cjllanwarne: Is there value in separating out the case f: Failure(t: ThrowableWithErrors) => f.errors ?
With appropriate FP magic to make the types sync up, of course

src/main/scala/wdl4s/WdlNamespace.scala
```diff
@@ -126,8 +128,9 @@ case class NamespaceWithWorkflow(importedAs: Option[String],
     if (failures.isEmpty) {
       Success(successes.map {case (k,v) => k -> v.get})
     } else {
-      val message = failures.values.collect {case f: Failure[_] => f.exception.getMessage}.mkString("\n")
-      Failure(new UnsatisfiedInputsException(s"Could not evaluate some declaration expressions:\n\n$message"))
+      val errors = failures.values.collect { case f: Failure[_] => f.exception.getMessage }
```
cjllanwarne: Same question here

cjllanwarne:
Can I make a potentially very unpopular comment...

I think these exceptions work better as common code (i.e. moved to Lenthall, any references to WDL excised)

src/main/scala/wdl4s/WdlExceptions.scala
Horneth: Not really - it was already there so I assumed it made sense to whoever wrote it.

Horneth:
I thought about that but wdl4s doesn't depend on lenthall, and I'm not convinced it should if we really want it to be an out of the box scala bindings for wdl.

geoffjentry:
@Horneth That doesn't preclude wdl4s to depend on lenthall. OTOH I don't feel strongly about @cjllanwarne 's suggestion either.

Horneth:
My main concern is that lenthall feels like it has a strong "Broad/Cromwell-ish" stamp - which we might not necessarily want to tie to a "standard WDL Scala library"

geoffjentry:
I really just view it as stuff common for our team's various codebases, not cromwell

cjllanwarne:
I like the idea of a class of exceptions which don't contain "WDL" in the namespace so that the CWL code could throw the same flavour of UnsatisfiedInputExceptions.

cjllanwarne:
Hmm, come to think of it, the CWL Java API probably throws CWL-specific exceptions so I think I've changed my mind.

geoffjentry:
Also agree. Otoh to play devils advocate this is such a simple exception class that I don't honk it's a big deal if it ends up duplicated in a few places (eg we could probably find this exact signature in multiple libraries were importing)

src/main/scala/wdl4s/WdlExceptions.scala
cjllanwarne: IIRC Java handles runtime exceptions differently to regular exceptions. But this doesn't feel like it should be one of those special cases - it looks to me like a genuine, predictable, handle-able exception.

geoffjentry:
Currently the cwl Java API is autogenerated so it's quite possible there's nothing of the sort.

src/main/scala/wdl4s/WdlExceptions.scala
cjllanwarne:
```
Runtime exceptions represent problems that are the result of a programming problem, and as such, the API client code cannot reasonably be expected to recover from them or to handle them in any way.
```

cjllanwarne:
Apart from thinking that RuntimeException is inappropriate here, :+1: from me.

src/main/scala/wdl4s/WdlExceptions.scala
Horneth: @cjllanwarne Would a simple `Throwable` be ok then ?

src/main/scala/wdl4s/WdlNamespace.scala
Horneth: In this particular case I don't think so because the only kind of exception possibly thrown during coercion is `UnsatisfiedInputException` which is not a `ThrowableWithErrors`.

src/main/scala/wdl4s/WdlExceptions.scala
cjllanwarne: Just extend Exception instead of RuntimeException IMO

src/main/scala/wdl4s/WdlExceptions.scala
```diff
@@ -0,0 +1,15 @@
...
+  def message: String
+  def errors: NonEmptyList[String]
+
+  override def getMessage = {
+    s"""$message\n${errors.list.mkString("\n")}"""
+  }
+}
+
+class UnsatisfiedInputsException(message: String) extends Exception(message)
+class ValidationException(val message: String, val errors: NonEmptyList[String]) extends Exception(message) with ThrowableWithErrors
```
geoffjentry: Why extend both exception and throwablewitherrors since the latter extends throwable?

Is there a case where you want the latter to not be an exception, and if not why not just make it an exception w/ errors? 

It's not a big deal either way though

geoffjentry:
Minor comment I don't care much about and :+1:

PR #5, opened 2016-01-14
Add new syntax error for when command references variable that doesn't exist

scottfrazer:
This WDL file:

```
task a {
  Int x
  command { ./script ${x+y} }
}

workflow w {
  call a
}
```

Gives this syntax error:

```
ERROR: Variable does not reference any declaration in the task (line 3, col 26):

  command { ./script ${x+y} }
                         ^

Task defined here (line 1, col 6):

task a {
     ^
```

This WDL:

```
task a {
  Int x
  Int y = x + z
  command { ./script ${x} }
}

workflow w {
  call a {input: x=5}
}
```

Will produce this error:

```
ERROR: Variable does not reference any declaration in the task (line 3, col 15):

  Int y = x + z
              ^

Declaration starts here (line 3, col 7):

  Int y = x + z
      ^
```

### DIFF
--- a/src/main/scala/wdl4s/AstTools.scala
+++ b/src/main/scala/wdl4s/AstTools.scala
@@ -18,7 +18,12 @@ object AstTools {
       AstTools.findAstsWithTrail(astNode, name, trail)
     }
     def findTerminalsWithTrail(terminalType: String, trail: Seq[AstNode] = Seq.empty): Map[Terminal, Seq[AstNode]] = {
-      AstTools.findTerminalsWithTrail(astNode, terminalType, trail)
+      astNode match {
+        case a: Ast => a.getAttributes.values.asScala flatMap { _.findTerminalsWithTrail(terminalType, trail :+ a) } toMap
+        case a: AstList => a.asScala.toVector flatMap { _.findTerminalsWithTrail(terminalType, trail :+ a) } toMap
+        case t: Terminal if t.getTerminalStr == terminalType => Map(t -> trail)
+        case _ => Map.empty[Terminal, Seq[AstNode]]
+      }
     }
     def findTopLevelMemberAccesses(): Iterable[Ast] = AstTools.findTopLevelMemberAccesses(astNode)
     def sourceString: String = astNode.asInstanceOf[Terminal].getSourceString
@@ -160,15 +165,6 @@ object AstTools {
       case x: AstList => x.asScala.toVector.flatMap{_.findAstsWithTrail(name, trail :+ x)}.toMap
       case x: Terminal => Map.empty[Ast, Seq[AstNode]]
       case _ => Map.empty[Ast, Seq[AstNode]]
-    }
-  }
-
-  def findTerminalsWithTrail(root: AstNode, name: String, trail: Seq[AstNode] = Seq.empty): Map[Terminal, Seq[AstNode]] = {
-    root match {
-      case a: Ast => a.getAttributes.values.asScala flatMap { _.findTerminalsWithTrail(name, trail :+ a) } toMap
-      case a: AstList => a.asScala.toVector flatMap { _.findTerminalsWithTrail(name, trail :+ a) } toMap
-      case t: Terminal if t.getTerminalStr == name => Map(t -> trail)
-      case _ => Map.empty[Terminal, Seq[AstNode]]
     }
   }
 
--- a/src/main/scala/wdl4s/AstTools.scala
+++ b/src/main/scala/wdl4s/AstTools.scala
@@ -1,6 +1,7 @@
 package wdl4s
 
 import java.io.File
+import scala.language.postfixOps
 
 import wdl4s.types._
 import wdl4s.values._
@@ -13,10 +14,10 @@ import scala.collection.JavaConverters._
 object AstTools {
   implicit class EnhancedAstNode(val astNode: AstNode) extends AnyVal {
     def findAsts(name: String): Seq[Ast] = AstTools.findAsts(astNode, name)
-    def findAstsWithTrail(name: String, trail: Seq[AstNode] = Seq.empty): Map[Ast, Seq[AstNode]] = {
+    def findAstsWithTrail(name: String, trail: Seq[AstNode] = Seq.empty): Map[Ast, Seq[AstNode]] =
       AstTools.findAstsWithTrail(astNode, name, trail)
-    }
-    def findTerminals(): Seq[Terminal] = AstTools.findTerminals(astNode)
+    def findTerminalsWithTrail(terminalType: String, trail: Seq[AstNode] = Seq.empty): Map[Terminal, Seq[AstNode]] =
+      AstTools.findTerminalsWithTrail(astNode, terminalType, trail)
     def findTopLevelMemberAccesses(): Iterable[Ast] = AstTools.findTopLevelMemberAccesses(astNode)
     def sourceString: String = astNode.asInstanceOf[Terminal].getSourceString
     def astListAsVector(): Seq[AstNode] = astNode.asInstanceOf[AstList].asScala.toVector
@@ -157,6 +158,15 @@ object AstTools {
       case x: AstList => x.asScala.toVector.flatMap{_.findAstsWithTrail(name, trail :+ x)}.toMap
       case x: Terminal => Map.empty[Ast, Seq[AstNode]]
       case _ => Map.empty[Ast, Seq[AstNode]]
+    }
+  }
+
+  def findTerminalsWithTrail(root: AstNode, name: String, trail: Seq[AstNode] = Seq.empty): Map[Terminal, Seq[AstNode]] = {
+    root match {
+      case a: Ast => a.getAttributes.values.asScala flatMap { _.findTerminalsWithTrail(name, trail :+ a) } toMap
+      case a: AstList => a.asScala.toVector flatMap { _.findTerminalsWithTrail(name, trail :+ a) } toMap
+      case t: Terminal if t.getTerminalStr == name => Map(t -> trail)
+      case _ => Map.empty[Terminal, Seq[AstNode]]
     }
   }
 
--- a/src/main/scala/wdl4s/AstTools.scala
+++ b/src/main/scala/wdl4s/AstTools.scala
@@ -14,10 +14,12 @@ import scala.collection.JavaConverters._
 object AstTools {
   implicit class EnhancedAstNode(val astNode: AstNode) extends AnyVal {
     def findAsts(name: String): Seq[Ast] = AstTools.findAsts(astNode, name)
-    def findAstsWithTrail(name: String, trail: Seq[AstNode] = Seq.empty): Map[Ast, Seq[AstNode]] =
+    def findAstsWithTrail(name: String, trail: Seq[AstNode] = Seq.empty): Map[Ast, Seq[AstNode]] = {
       AstTools.findAstsWithTrail(astNode, name, trail)
-    def findTerminalsWithTrail(terminalType: String, trail: Seq[AstNode] = Seq.empty): Map[Terminal, Seq[AstNode]] =
+    }
+    def findTerminalsWithTrail(terminalType: String, trail: Seq[AstNode] = Seq.empty): Map[Terminal, Seq[AstNode]] = {
       AstTools.findTerminalsWithTrail(astNode, terminalType, trail)
+    }
     def findTopLevelMemberAccesses(): Iterable[Ast] = AstTools.findTopLevelMemberAccesses(astNode)
     def sourceString: String = astNode.asInstanceOf[Terminal].getSourceString
     def astListAsVector(): Seq[AstNode] = astNode.asInstanceOf[AstList].asScala.toVector
--- a/src/main/scala/wdl4s/Task.scala
+++ b/src/main/scala/wdl4s/Task.scala
@@ -23,6 +23,7 @@ object Task {
     val runtimeAttributes = RuntimeAttributes(ast)
     val meta = wdlSectionToStringMap(ast, AstNodeName.Meta, wdlSyntaxErrorFormatter)
     val parameterMeta = wdlSectionToStringMap(ast, AstNodeName.ParameterMeta, wdlSyntaxErrorFormatter)
+    val outputs = ast.findAsts(AstNodeName.Output) map { TaskOutput(_, declarations, wdlSyntaxErrorFormatter) }
 
     if (commandAsts.size != 1) throw new SyntaxError(wdlSyntaxErrorFormatter.expectedExactlyOneCommandSectionPerTask(taskNameTerminal))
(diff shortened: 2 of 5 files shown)


scottfrazer:
WHEEL: @mcovarr @geoffjentry

scottfrazer:
@yfarjoun this one's for you :)

yfarjoun:
YEY!

cjllanwarne:
Does PyWDL have an equivalent change upcoming?

scottfrazer:
Yeah, PyWDL is missing a lot of these higher-level syntax checks actually

src/test/scala/wdl4s/SyntaxErrorSpec.scala
```diff
@@ -256,5 +256,17 @@ class SyntaxErrorSpec extends FlatSpec with Matchers {
...
   }
+  it should "detect when expressions within a command block don't reference actual variables" in {
+    expectError(
+      """task a {
+        |  Int x
+        |  command { ./script ${x+y} }
+        |}
+        |
+        |workflow w {
+        |  call a
```
cjllanwarne: From a test-pedantry point of view, this could just be throwing an error because "x" is never set. To be exact, make sure you say `call a {input: x=5 }` or equivalent

cjllanwarne: Actually, does assigning to a variable that's never declared also constitute a syntax error?

scottfrazer: I believe `call a {input: x=5, y=5 }` ought to return an syntax error in this case.  If that syntax error is not already present I'll add it.

cjllanwarne:
This checks for unfound declarations in the command text. Does it also validate missing declarations in other declarations? e.g.
```
task X {
  Int x
  Int z = x + y
}
```

scottfrazer:
@cjllanwarne I'll check on that too.  The algorithm should be the same to check those expressions so I'll extend it to also cover those expressions.

src/test/scala/wdl4s/SyntaxErrorSpec.scala
scottfrazer: I just confirmed that there is indeed a syntax error and test covering this already.

scottfrazer:
@cjllanwarne I added the same kind of syntax check for declarations in tasks... see the description for what that error message looks like.

src/main/scala/wdl4s/WdlSyntaxErrorFormatter.scala
```diff
@@ -204,4 +204,39 @@ case class WdlSyntaxErrorFormatter(terminalMap: Map[Terminal, WdlSource]) extend
...
+        |
+        |${pointToSource(variable)}
+        |
+        |Declaration starts here (line ${declarationName.getLine}, col ${declarationName.getColumn}):
+        |
+        |${pointToSource(declarationName)}
+     """.stripMargin
+  }
+
+  def variableDeclaredMultipleTimes(first: Terminal, second: Terminal) = {
```
cjllanwarne: Ooh, are we also covered for DSDEEPB-2531 now?

cjllanwarne: Looking at the ticket, there's 3 test cases. I think this covers the first?

scottfrazer: I will look

src/main/scala/wdl4s/Task.scala
```diff
@@ -16,15 +16,48 @@ import scala.util.Try
...
+        if (!declarations.map(_.name).contains(variable.getSourceString)) {
+          throw new SyntaxError(wdlSyntaxErrorFormatter.commandExpressionContainsInvalidVariableReference(nameTerminal, variable))
+        }
+      }
+    }
+
+    def findDeclarationAsts(decl: Declaration) =
+      ast.findAsts(AstNodeName.Declaration) collect { case a: Ast if a.getAttribute("name").sourceString == decl.name => a }
+
+    declarations.foldLeft(Seq.empty[Declaration]) { (acc, cur) =>
```
cjllanwarne: This foldLeft function is a little unwieldy in-line. Can you break it out into a separate def?

src/main/scala/wdl4s/Task.scala
```diff
@@ -30,6 +30,38 @@ object Task {
...
+      cur.expression match {
+        case Some(e) => e.ast.findTerminals("identifier") foreach {variable =>
+          if (!acc.map(_.name).contains(variable.getSourceString)) {
+            // .head below because we are assuming if you have a Declaration object that it must have come from a Declaration AST
+            val declarationAst = findDeclarationAsts(cur).head
+            throw new SyntaxError(wdlSyntaxErrorFormatter.declarationContainsInvalidVariableReference(declarationAst.getAttribute("name").asInstanceOf[Terminal], variable))
+          }
+        }
+        case None =>
+      }
```
mcovarr: converting the `match` to a `foreach` would eliminate the pattern match on `Some` and the need to have a `None` cass

src/main/scala/wdl4s/WdlSyntaxErrorFormatter.scala
mcovarr: I didn't see a test for this

src/main/scala/wdl4s/Task.scala
```diff
@@ -30,6 +30,38 @@ object Task {
       case x: Ast => ParameterCommandPart(x, wdlSyntaxErrorFormatter)
     }
 
+    commandTemplate collect { case x: ParameterCommandPart => x } foreach { p =>
+      p.expression.ast.findTerminals("identifier") foreach {variable =>
```
scottfrazer: this `findTerminals("identifier")` is too permissive.  It also catches function names... I need to fix that.

src/test/scala/wdl4s/SyntaxErrorSpec.scala
```diff
@@ -260,9 +260,6 @@ class SyntaxErrorSpec extends FlatSpec with Matchers {
     expectError(
       """task a {
         |  command { ./script }
-        |  output {
-        |    Int x = "bad value"
-        |  }
```
mcovarr: Why the change to what appears to be an unrelated test?

src/main/scala/wdl4s/WdlSyntaxErrorFormatter.scala
```diff
@@ -207,22 +207,57 @@ case class WdlSyntaxErrorFormatter(terminalMap: Map[Terminal, WdlSource]) extend
...
+    s"""ERROR: Variable '${first.getSourceString}' is declared more than once.
+        |
+        |First occurrence (line ${first.getLine}, col ${first.getColumn}):
+        |
+        |${pointToSource(first)}
+        |
+        |Second occurence (line ${second.getLine}, col ${second.getColumn}):
```
mcovarr: At least one of the spellings of the word after 'First' and 'Second' is incorrect.  :smile:

src/main/scala/wdl4s/Task.scala
```diff
@@ -30,6 +30,38 @@ object Task {
...
+    def findDeclarationAsts(decl: Declaration) =
+      ast.findAsts(AstNodeName.Declaration) collect { case a: Ast if a.getAttribute("name").sourceString == decl.name => a }
+
+    declarations.foldLeft(Seq.empty[Declaration]) { (acc, cur) =>
+      findDeclarationAsts(cur) match {
+        case declAsts if declAsts.size > 1 =>
+          val declNameTerminals = declAsts.map(_.getAttribute("name").asInstanceOf[Terminal])
+          throw new SyntaxError(wdlSyntaxErrorFormatter.variableDeclaredMultipleTimes(declNameTerminals(0), declNameTerminals(1)))
+        case _ =>
+      }
```
mcovarr: could be a `collect` to get rid of the `case _` line

src/test/scala/wdl4s/SyntaxErrorSpec.scala
scottfrazer: I want to keep these WDL files to only contain the exact syntax error that they're actually testing for and nothing else.  In this case, the output section is not needed to test the `meta` section.  Worse than that though is that the output section itself contains a syntax error so it's actually masking the real syntax error of this test.  `expectError` really should probably match the error it gets back against a string so you know you got the right error.  I can file a ticket for that, because I don't want to go down too much of a rabbit hole

src/main/scala/wdl4s/Task.scala
scottfrazer: hmmm I'm not sure how exactly...

src/main/scala/wdl4s/WdlSyntaxErrorFormatter.scala
scottfrazer: I'll add it

src/main/scala/wdl4s/Task.scala
scottfrazer: Maybe it's better as an if statement?

mcovarr: sure, the collect I had in mind didn't actually make sense.  :smile:

src/test/scala/wdl4s/SyntaxErrorSpec.scala
```diff
@@ -285,5 +282,77 @@ class SyntaxErrorSpec extends FlatSpec with Matchers {
...
+      """task inputOutputOops {
+        |  Int c = 5
+        |  command { echo 5 }
+        |  output {
+        |    Int c = 10
+        |  }
+        |}
+        |workflow c { call inputOutputOops }
+      """.stripMargin)
+  }
```
scottfrazer: @cjllanwarne there's your DSDEEPB-2531 WDLs ^^^

scottfrazer:
@mcovarr @cjllanwarne I pushed all code review changes

src/main/scala/wdl4s/Task.scala
```diff
@@ -30,7 +31,87 @@ object Task {
...
+    val variablesReferencedInCommand = for {
+      param <- commandTemplate collect { case x: ParameterCommandPart => x }
+      variable <- param.expression.variableReferences
+    } yield variable
+
+    variablesReferencedInCommand foreach { variable =>
+      if (!declarations.map(_.name).contains(variable.getSourceString)) {
+        throw new SyntaxError(wdlSyntaxErrorFormatter.commandExpressionContainsInvalidVariableReference(taskNameTerminal, variable))
+      }
+    }
```
mcovarr: I think this would be better as `Validation`s that accumulate errors rather than failing fast on the first problem

src/main/scala/wdl4s/Task.scala
```diff
@@ -30,7 +31,87 @@ object Task {
...
+      if (declarationAstsWithSameName.nonEmpty)
+        throw new SyntaxError(wdlSyntaxErrorFormatter.variableDeclaredMultipleTimes(declNameTerminals(0), outputNameTerminals(0)))
+
+      current.expression.variableReferences foreach { variable =>
+        val varName = variable.getSourceString
+        if (!accumulated.map(_.name).contains(varName) && !declarations.map(_.name).contains(varName)) {
+          throw new SyntaxError(wdlSyntaxErrorFormatter.declarationContainsInvalidVariableReference(
+            taskOutputAstsWithSameName.head.getAttribute("var").asInstanceOf[Terminal],
+            variable
+          ))
```
mcovarr: More fast failing here

src/main/scala/wdl4s/AstTools.scala
```diff
@@ -13,10 +14,10 @@ import scala.collection.JavaConverters._
 object AstTools {
   implicit class EnhancedAstNode(val astNode: AstNode) extends AnyVal {
     def findAsts(name: String): Seq[Ast] = AstTools.findAsts(astNode, name)
-    def findAstsWithTrail(name: String, trail: Seq[AstNode] = Seq.empty): Map[Ast, Seq[AstNode]] = {
+    def findAstsWithTrail(name: String, trail: Seq[AstNode] = Seq.empty): Map[Ast, Seq[AstNode]] =
       AstTools.findAstsWithTrail(astNode, name, trail)
-    }
-    def findTerminals(): Seq[Terminal] = AstTools.findTerminals(astNode)
+    def findTerminalsWithTrail(terminalType: String, trail: Seq[AstNode] = Seq.empty): Map[Terminal, Seq[AstNode]] =
+      AstTools.findTerminalsWithTrail(astNode, terminalType, trail)
```
mcovarr: I believe our standards say to use curlies if it's not a one liner

src/main/scala/wdl4s/Task.scala
```diff
@@ -30,7 +31,87 @@ object Task {
...
+    }
+
+    /**
+      * Ensures that the current declaration doesn't have a name conflict with another declaration
+      * and that the expression for the current declaration only has valid variable references in it
+      *
+      * @param accumulated The declarations that come lexically before 'current'
+      * @param current The declaration being validated
+      */
+    def validateDeclaration(accumulated: Seq[Declaration], current: Declaration): Seq[Declaration] = {
```
mcovarr: and the way this is used will also result in fast failing on the first bad declaration

src/main/scala/wdl4s/Task.scala
scottfrazer: What do you think now?

mcovarr: Much better, thanks!  I would suggesting adding a note as to why converting this to `Validation`s is not straightforward.  I'm also not sure if there's any value in putting the errors in a `SyntaxError` box; these could perhaps just be `String`s.

scottfrazer: Okay, will do

scottfrazer: Updated

src/main/scala/wdl4s/AstTools.scala
```diff
@@ -16,7 +17,9 @@ object AstTools {
     def findAstsWithTrail(name: String, trail: Seq[AstNode] = Seq.empty): Map[Ast, Seq[AstNode]] = {
       AstTools.findAstsWithTrail(astNode, name, trail)
     }
-    def findTerminals(): Seq[Terminal] = AstTools.findTerminals(astNode)
+    def findTerminalsWithTrail(terminalType: String, trail: Seq[AstNode] = Seq.empty): Map[Terminal, Seq[AstNode]] = {
```
geoffjentry: These aliases were originally just to minimize the impact of some changes I made a while back. Since this is a new one it might as well be fully and completely be in here instead of a passthrough.

mcovarr:
:+1:

src/main/scala/wdl4s/AstTools.scala
scottfrazer: Okay it's moved

geoffjentry:
Glam

geoffjentry:
Ugh. Lgtm

PR #6, opened 2016-01-14
findCallsByName

scottfrazer:


cjllanwarne:
The more I see of this, the more convinced I am that winstanley should be calling into it. This is exactly the type of helper method the tutorial creates for "find-usages", "auto-completion", etc etc.

cjllanwarne:
:+1:

src/main/scala/wdl4s/Workflow.scala
```diff
@@ -133,6 +133,8 @@ case class Workflow(unqualifiedName: String,
     (callInputs ++ declarationInputs) map { input => input.fqn -> input } toMap
   }
 
+  def findCallByName(name: String): Option[Call] = calls.find(_.unqualifiedName == name)
```
mcovarr: doc that the param is an lqn

mcovarr:
:+1:

PR #7, opened 2016-01-20
Scoped declarations

scottfrazer:
This is part of DSDEEPB-2455

`Declaration` previously was not scoped, but we tried to fake it by passing it a FQN string.  However, we were passing the literal string `"name"` as the FQN for all declarations that were coming from tasks.  This is leading Cromwell to storing `name` as scope FQNs in the database.

This patch allows for `Workflow` and `Call` to call `.scopedDeclarations` to get a declaration that's properly scoped, and can provide a real implementation of `.fullyQualifiedName`

Also changed:

* `TaskOutput` is now a `Declaration` that has a required expression
* All declarations on the task level are type checked at compile time
* Removed `inputs` field on `Task`. Using `declarations` field instead.

### DIFF
--- a/src/main/scala/wdl4s/Call.scala
+++ b/src/main/scala/wdl4s/Call.scala
@@ -27,7 +27,7 @@ object Call {
     val callInputSectionMappings = processCallInput(ast, wdlSyntaxErrorFormatter)
 
     callInputSectionMappings foreach { case (taskParamName, expression) =>
-      task.inputs find { taskInput => taskInput.name == taskParamName } getOrElse {
+      task.declarations find { decl => decl.name == taskParamName } getOrElse {
         /*
           FIXME-ish
           It took me a while to figure out why this next part is necessary and it's kind of hokey.
@@ -92,7 +92,7 @@ case class Call(alias: Option[String],
    * are satisfied via the 'input' section of the Call definition.
    */
   def unsatisfiedInputs: Seq[WorkflowInput] = for {
-    i <- task.inputs if !inputMappings.contains(i.name)
+    i <- task.declarations if !inputMappings.contains(i.name) && i.expression.isEmpty
   } yield WorkflowInput(s"$fullyQualifiedName.${i.name}", i.wdlType, i.postfixQuantifier)
 
   override def toString: String = s"[Call name=$unqualifiedName, task=$task]"
@@ -102,6 +102,11 @@ case class Call(alias: Option[String],
    */
   def instantiateCommandLine(inputs: CallInputs, functions: WdlFunctions[WdlValue]): Try[String] =
     task.instantiateCommand(inputs, functions)
+
+  /**
+    * @return Seq[ScopedDeclaration] which are scoped to this Call
+    */
+  def scopedDeclarations: Seq[ScopedDeclaration] = task.declarations.map(decl => ScopedDeclaration(this, decl))
 
   override def rootWorkflow: Workflow = parent map { _.rootWorkflow } getOrElse { throw new IllegalStateException("Call not in workflow") }
 }
--- a/src/main/scala/wdl4s/Declaration.scala
+++ b/src/main/scala/wdl4s/Declaration.scala
@@ -5,9 +5,8 @@ import wdl4s.types.WdlType
 import wdl4s.parser.WdlParser.{Ast, AstNode}
 
 object Declaration {
-  def apply(ast: Ast, scopeFqn: FullyQualifiedName, wdlSyntaxErrorFormatter: WdlSyntaxErrorFormatter): Declaration = {
-    Declaration(
-      scopeFqn,
+  def apply(ast: Ast, wdlSyntaxErrorFormatter: WdlSyntaxErrorFormatter): Declaration = {
+    UnscopedDeclaration(
       ast.getAttribute("type").wdlType(wdlSyntaxErrorFormatter),
       Option(ast.getAttribute("postfix")).map(_.sourceString),
       ast.getAttribute("name").sourceString,
@@ -34,16 +33,34 @@ object Declaration {
  *
  * Both the definition of test_file and wf_string are declarations
  */
-case class Declaration(scopeFqn: FullyQualifiedName, wdlType: WdlType, postfixQuantifier: Option[String], name: String, expression: Option[WdlExpression]) {
+trait Declaration {
+  def wdlType: WdlType
+  def postfixQuantifier: Option[String]
+  def name: String
+  def expression: Option[WdlExpression]
+  def asTaskInput: Option[TaskInput] = expression match {
+    case Some(expr) => None
+    case None => Option(TaskInput(name, wdlType, postfixQuantifier))
+  }
+
+  def toWdlString: String = {
+    val expr = expression.map(e => s" = ${e.toWdlString}").getOrElse("")
+    s"${wdlType.toWdlString} $name$expr"
+  }
+}
+
+case class UnscopedDeclaration(wdlType: WdlType, postfixQuantifier: Option[String], name: String, expression: Option[WdlExpression]) extends Declaration
+
+object ScopedDeclaration {
+  def apply(scope: Scope, decl: Declaration): ScopedDeclaration = {
+    ScopedDeclaration(scope, decl.wdlType, decl.postfixQuantifier, decl.name, decl.expression)
+  }
+}
+
+case class ScopedDeclaration(scope: Scope, wdlType: WdlType, postfixQuantifier: Option[String], name: String, expression: Option[WdlExpression]) extends Declaration {
+  def fullyQualifiedName: FullyQualifiedName = s"${scope.fullyQualifiedName}.$name"
   def asWorkflowInput: Option[WorkflowInput] = expression match {
     case Some(expr) => None
     case None => Some(WorkflowInput(fullyQualifiedName, wdlType, postfixQuantifier))
   }
-
-  def asTaskInput: Option[TaskInput] = expression match {
-    case Some(expr) => None
-    case None => Some(TaskInput(name, wdlType, postfixQuantifier))
-  }
-
-  def fullyQualifiedName: FullyQualifiedName = s"$scopeFqn.$name"
 }
--- a/src/main/scala/wdl4s/Task.scala
+++ b/src/main/scala/wdl4s/Task.scala
@@ -4,26 +4,37 @@ import java.util.regex.Pattern
 
 import wdl4s.AstTools.{AstNodeName, EnhancedAstNode}
 import wdl4s.command.{CommandPart, ParameterCommandPart, StringCommandPart}
-import wdl4s.expression.WdlFunctions
+import wdl4s.expression.{NoFunctions, WdlStandardLibraryFunctionsType, WdlFunctions}
 import wdl4s.parser.WdlParser._
+import wdl4s.types.{WdlIntegerType, WdlType}
 import wdl4s.values.WdlValue
 
 import scala.annotation.tailrec
 import scala.collection.JavaConverters._
 import scala.language.postfixOps
-import scala.util.Try
+import scala.util.{Failure, Success, Try}
 
 object Task {
   val Ws = Pattern.compile("[\\ \\t]+")
+
+  /** The function validateDeclaration() and the DeclarationAccumulator class are used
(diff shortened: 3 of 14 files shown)


kshakir:
:+1:

scottfrazer:
@kshakir and wheel chooses @cjllanwarne

src/main/scala/wdl4s/Declaration.scala
```diff
@@ -34,16 +33,33 @@ object Declaration {
...
+trait Declaration {
+  def wdlType: WdlType
+  def postfixQuantifier: Option[String]
+  def name: String
+  def expression: Option[WdlExpression]
   def asTaskInput: Option[TaskInput] = expression match {
     case Some(expr) => None
-    case None => Some(TaskInput(name, wdlType, postfixQuantifier))
+    case None => Option(TaskInput(name, wdlType, postfixQuantifier))
+  }
```
geoffjentry: \n

src/main/scala/wdl4s/WdlNamespace.scala
```diff
@@ -111,28 +112,24 @@ case class NamespaceWithWorkflow(importedAs: Option[String],
...
-    if (failures.isEmpty) {
-      Success(successes.map {case (k,v) => k -> v.get})
-    } else {
-      val errors = failures.values.collect { case f: Failure[_] => f.exception.getMessage }
-        // .get because failures is guaranteed to be nonEmpty
-      Failure(new ValidationException("Could not evaluate some declaration expressions", errors.toList.toNel.get))
+    def evalDeclaration(accumulated: Map[String, Try[WdlValue]], current: ScopedDeclaration): Map[String, Try[WdlValue]] = {
+      current.expression match {
+        case Some(expr) =>
+          val successfulAccumulated = accumulated.collect({case (k, v) if v.isSuccess => k -> v.get })
```
geoffjentry: space after {

src/main/scala/wdl4s/WdlNamespace.scala
```diff
@@ -111,28 +112,24 @@ case class NamespaceWithWorkflow(importedAs: Option[String],
...
+      }
     }
+
+    val workflowStaticDeclarations = workflow.scopedDeclarations.foldLeft(Map.empty[String, Try[WdlValue]])(evalDeclaration)
+
+    val callStaticDeclarations = workflow.calls flatMap { call =>
+      call.scopedDeclarations.foldLeft(Map.empty[String, Try[WdlValue]])(evalDeclaration)
+    } toMap
+
+    TryUtil.sequenceMap(workflowStaticDeclarations ++ callStaticDeclarations)
```
geoffjentry: Why accumulate trys instead of a validation?

scottfrazer: I wanted to stick with the standard library.  Any opinions @kshakir or @cjllanwarne?

src/main/scala/wdl4s/Task.scala
```diff
@@ -4,26 +4,37 @@ import java.util.regex.Pattern
```
scottfrazer: Sorry for the bad diff on this file...

* Before there were two fold functions (one for checking the input `Declaration`s and one for checking output `TaskOutput`s)
* I made `TaskOutput` into a `Declaration` so now there needs to be only one fold function.
* I added `typeCheckExpression()` for checking type compatibility for all `Declaration`s
* Removed `inputs` field on `Task`.  Using `declarations` field instead.

src/main/scala/wdl4s/WdlNamespace.scala
geoffjentry: As a general rule of thumb across all of our code we should be moving away for our homebrew scheme which was only used because we weren't already using validations. Now that we are, there's a far more idiomatic way to express accumulation of badness.

scottfrazer: I'd rather talk about the merits of each approach rather than just making assertions.  Arguments about idioms are hard to verify, can you point me to some resources arguing for *why* Validations should be idiomatic?  Or at least code examples that aren't in Cromwell (I've probably read a majority of the tutorials that are easily found from Google searches)?  My experience with Validations has been nothing but pain and frustration.  Being told "use it because it's idiomatic" doesn't help.

What better way to assess relative merits than a side-by-side comparison!  After all, I want to be convinced and I want to agree with you so let me write it up both ways, make a Gist that shows them side-by-side.  I'm struggling with my implementation using Validations right now so I could use some help from either you or @mcovarr

kshakir: I'm on the fence and will abstain from the vote leaving only comments:

I believe that `scalaz.Validation`, in the absense of another validation library, is the direction our wdl/cromwell codebase should be going.

But, I've also banged my head sometimes coming up with the proper way to use the cumbersome library (for example, trying to run two validations on the name value), poor easy to find or understand code samples, and little documentation written in plain english. That said, with light digging, I did find a page on Safari Books written in [english](http://techbus.safaribooksonline.com/9781491950135/optionsandothercontainertypes_html#ix_scalazval), and not mathematical set theory (I'm looking at you eed3si9n), so maybe there's more out there?

scottfrazer: > I've also banged my head sometimes coming up with the proper way to use the cumbersome library

I hear ya.  I'm currently in the head-banging mode right now.  I'll see if I can get access to that book.  I'm surprised there aren't better online resources.

So basically right now I'm stuck.  I can sink a hours or a day into figuring out how to make this work with scalaz, but unless somebody wants to sit side-by-side, I'm going to give up in about 15 more minutes.

Here's some work-in-progress code: https://gist.github.com/scottfrazer/25781962cc85276b5bd9

cjllanwarne: Since you asked... my two cents is that scalaz makes everything it touches slightly terser but completely unreadable or understandable. And people are much less likely to want to maintain or extend it. 

If we can wrap the clever stuff into a human-readable library or have a 1-page cheat sheet than I'm all for it. Otherwise I'm afraid it gets a down-vote :-1: from me.

geoffjentry: While I generally agree with you the validation stuff isn't terser and is really the same thing. A Try is effectively an Either[Throwable, A]. A Validation is effectively* an Either[A, B]. TryUtil is taking those Either[Throwable, A], unboxing the messages out of the Throwable and combining them. The Validation applicative is combining the As. 

* Validation doesn't have flatMap**
** It does, but it's deprecated

My stance is that if we're reinventing a wheel the onus is on us to show that it is demonstrably superior to the tool that's more generally used. I don't see how TryUtil is superior in any fashion - I don't see how it's easier to comprehend and it's inherently less inefficient due to the unboxing.

That said I don't care enough here and now to argue this specific case but I view just about everything inside of TryUtil as a code smell, sequence included

kshakir: - Info on getting access to Safari Books is [here](https://iwww.broadinstitute.org/bits/additional-it-perks).
- Re: deprecated flatMap-- yet another reason the scalaz community needs _clear_ online examples. The Safari example I linked to is very blatantly simplifying `scalaz.Validation` by using for-comprehensions.

scottfrazer: Hey guess what!  We have a side-by-side comparison!

https://gist.github.com/scottfrazer/25781962cc85276b5bd9

Let me know if you have any obvious ways that the scalaz version can be improved and I'll incorporate them.

* Scalaz version is way more difficult for me.  Thanks @mcovarr for helping put it together though
* IntelliJ was painfully slow trying to figure out all the scalaz stuff.
* No obvious way to convert a `Try` to a `ValidationNel[String, WdlValue]`.  Anybody have any ideas?
* `attemptedEvaluations.foldLeft(Map.empty[String, WdlValue].successNel[String])((acc, v) => (acc |@| v._2) { (a, v0) => a + (v._1 -> v0) })` is basically the scalaz equivalent to `TryUtil.sequenceMap`
* Returning `ValidationNel` has a spidering effect.  All client code now needs to be able to deal with these things.

> My stance is that if we're reinventing a wheel the onus is on us to show that it is demonstrably superior to the tool that's more generally used.

I agree

>  I don't see how TryUtil is superior in any fashion.

Can you tell me then why you think the scalaz is superior in this case?

geoffjentry: @kshakir It irritates me as well. My inclination is often to map over them and IntelliJ doesn't alert me that I'm going to get barked at when I do my sbt assembly. Apparently it really got on some folks' nerves that flatMap was there despite not being a monad. The counterargument was "It's been there for ages and is way convenient" but request denied.

geoffjentry: @scottfrazer 

* I don't view one as being more arcane than the other. I think you're somewhat biased in that you already know the wizardry of one and not the other. If someone knew neither IMO they'd both look freaking weird.
* I have been surprised myself that they don't have a convenience implicit (at least not that I've found) for converting - I still don't fully believe it doesn't exist but I'll admit I've come across the same thing. IMO *that* is the sort of thing which should be in TryUtil. 
* Yes it is. My complaint about our homebrew scheme is that it's effectively equivalent plus some extra unboxing and doesn't provide any extra benefit AFAICT. That doesn't scream to me "please let's homebrew our own thing". So I'm not stating superiority, I'm saying that the tryutil scheme is *at best* equal and we shouldn't be blazing our own trail in cases where we're not providing a demonstrable value add.
* It spiders to exactly the same extent that Try does, which is to say as far as you want it to go. With the exception of "how do you operate over multiple of these" there really is no difference at all between how you handle a validation vs a try.

The one place where I think dealing with the validations is a bit of a headache is when you want to combine fail-slow with fail-fast. However it's still no different than what you'd have in the tryutil model, it's just different syntactically. The headache has been remembering what's intended to be fast vs slow and not combining them. At least in this model the two bits are differently typed (a validation vs a disjunction) so it's harder to screw it up w/o realizing it.

src/main/scala/wdl4s/Call.scala
```diff
@@ -92,7 +92,8 @@ case class Call(alias: Option[String],
    * are satisfied via the 'input' section of the Call definition.
    */
   def unsatisfiedInputs: Seq[WorkflowInput] = for {
-    i <- task.inputs if !inputMappings.contains(i.name)
+    i <- task.declarations
+    if !inputMappings.contains(i.name) && i.expression.isEmpty
```
cjllanwarne: strange newline?

src/main/scala/wdl4s/WdlNamespace.scala
```diff
@@ -111,28 +111,20 @@ case class NamespaceWithWorkflow(importedAs: Option[String],
...
-    val collected = mutable.Map[String, WdlValue]()
-    val allDeclarations = workflow.declarations ++ workflow.calls.flatMap {_.task.declarations}
-
-    val evaluatedDeclarations = allDeclarations.filter {_.expression.isDefined}.map {decl =>
-      val value = decl.expression.get.evaluate(declarationLookupFunction(decl, collected.toMap ++ userInputs), wdlFunctions)
-      collected += (decl.fullyQualifiedName -> value.get)
-      val coercedValue = value match {
-        case Success(s) => decl.wdlType.coerceRawValue(s)
-        case f => f
+    def evalDeclaration(accumulated: Map[String, Try[WdlValue]], current: ScopedDeclaration): Map[String, Try[WdlValue]] = {
```
cjllanwarne: accumulator + transform function = fold?

src/main/scala/wdl4s/Workflow.scala
```diff
@@ -129,7 +129,7 @@ case class Workflow(unqualifiedName: String,
    */
   def inputs: Map[FullyQualifiedName, WorkflowInput] = {
     val callInputs = for { call <- calls; input <- call.unsatisfiedInputs } yield input
-    val declarationInputs = for { declaration <- declarations; input <- declaration.asWorkflowInput } yield input
+    val declarationInputs = for { declaration <- scopedDeclarations; input <- declaration.asWorkflowInput } yield input
```
cjllanwarne: Here newlines would be good!

cjllanwarne:
:+1:

PR #8, opened 2016-01-25
Runtime expressions

scottfrazer:


src/main/java/wdl4s/parser/MemoryUnit.java
```diff
@@ -0,0 +1,22 @@
+package wdl4s.parser;
+
+public enum MemoryUnit {
```
cjllanwarne: is this because scala can't do enums with values? Isn't this just a case class?

src/main/scala/wdl4s/MemorySize.scala
```diff
@@ -0,0 +1,39 @@
...
+
+import wdl4s.parser.MemoryUnit
+
+import scala.language.postfixOps
+import scala.util.{Failure, Success, Try}
+import scalaz.Scalaz._
+
+object MemorySize {
+  val memoryPattern = """(\d+(?:\.\d+)?)\s*(\w+)""".r
+  def parse(unparsed: String): Try[MemorySize] = {
```
cjllanwarne: Is this a case where it's odd to be converting `Validation`s back into `Try`s?

cjllanwarne:
Under the new tyranny (long live it), should we wait until both repos are ready for review, then review both in synchrony?

src/main/scala/wdl4s/MemorySize.scala
geoffjentry: as per what I was saying in person - I think the point at which one says "This is a full and complete unit of error checking, after this is will no longer be combined with other errors", that's the point to either convert to a Try or throw the exception.

Just one man's opinion

src/main/java/wdl4s/parser/MemoryUnit.java
scottfrazer: I *think* so?  There was an enum here prior to this called `MemorySize`, so I just kept with the same paradigm of defining the enum in Java.

scottfrazer:
Yeah, I wouldn't waste too much time reviewing this now... sorry I forgot that I shouldn't have even filed this PR.  Habits can be hard to break.

PR #9, opened 2016-02-03
Runtime expressions

scottfrazer:
Corresponding Cromwell change: https://github.com/broadinstitute/cromwell/pull/420
WDL spec updates: https://github.com/broadinstitute/wdl/pull/27


src/main/java/wdl4s/parser/MemoryUnit.java
```diff
@@ -0,0 +1,22 @@
+package wdl4s.parser;
+
+public enum MemoryUnit {
+    Bytes(1, "B"),
+    KiB(1 << 10, "KiB"),
+    MiB(1 << 20, "MiB"),
+    GiB(1 << 30, "GiB"),
+    TiB(1099511627776.0, "TiB"), // 1 << 40
```
cjllanwarne: One of these things is not like the others

src/main/scala/wdl4s/MemorySize.scala
```diff
@@ -0,0 +1,39 @@
...
+
+import wdl4s.parser.MemoryUnit
+
+import scala.language.postfixOps
+import scala.util.{Failure, Success, Try}
+import scalaz.Scalaz._
+
+object MemorySize {
+  val memoryPattern = """(\d+(?:\.\d+)?)\s*(\w+)""".r
+  def parse(unparsed: String): Try[MemorySize] = {
```
cjllanwarne: I think ErrorOr[MemorySize] is a better return type here since it's all Validations internally anyway

src/main/scala/wdl4s/RuntimeAttributes.scala
```diff
@@ -1,76 +1,91 @@
...
-  /**
-    * Checks that a provided memory value is valid, if not throws an IllegalArgumentException
-    */
-  private def validateMemory(value: String): Unit = {
-    validateMemStringInGb (value) match {
-      case Failure(f) => throw new IllegalArgumentException("Invalid memory value specified:\n" + f.toList.mkString("\n"))
-      case Success(_) => ()
+  def validateMemoryValue(value: WdlValue): Try[MemorySize] = {
+    value match {
+      case i: WdlInteger => Success(MemorySize(i.value.toDouble, MemoryUnit.GB))
```
cjllanwarne: It seems odd and potentially counter-intuitive that "GB" is the default.

I can imagine you might write something like `memory = file_size(input_file)` and accidentally getting 10^9 more memory than you expected!

cjllanwarne: Also, if this MUST be GB, then at least allow floats here too!

src/main/scala/wdl4s/MemorySize.scala
```diff
@@ -0,0 +1,39 @@
...
+}
+
+case class MemorySize(amount: Double, unit: MemoryUnit) {
+  def bytes: Double = amount * unit.bytes
+  def to(unit: MemoryUnit): MemorySize = MemorySize(this.bytes / unit.bytes, unit)
+  override def toString: String = {
+    val adjustedAmount = (unit, amount) match {
+      case (MemoryUnit.Bytes, a) => a.ceil.toLong.toString
+      case (_, a) if a == a.toLong => a.toLong.toString
+      case (_, a) => a.toString
```
cjllanwarne: Update the WDL spec - it currently says this has to be an Integer value, but this is clearly designed to allow (e.g.) 3.5 GB

src/main/scala/wdl4s/RuntimeAttributes.scala
mcovarr: The existing behavior is to default to GB and this is what the current pipeline expects.

src/main/scala/wdl4s/values/WdlBoolean.scala
```diff
@@ -8,6 +8,7 @@ object WdlBoolean {
   val True = new WdlBoolean(true)
   val False = new WdlBoolean(false)
   def apply(value: Boolean) = if (value) WdlBoolean.True else WdlBoolean.False
+  def unapply(b: WdlBoolean): Option[Boolean] = Option(b.value)
```
cjllanwarne: Wait, what?

cjllanwarne:
One more test suggestion w.r.t inputs using other inputs - how about:

```
task x {
  Int a
  Int b = a * 1000
  String c = b + "GB"
  ...
  runtime {
    memory: c
  }
}
```

src/main/scala/wdl4s/RuntimeAttributes.scala
cjllanwarne: TBH I'd rather this default didn't exist at all (it isn't in the WDL spec). The current pipeline specifies a unit for every `memory: _` value.

mcovarr: oops ignore me, I was confusing memory with disk :frowning:

src/main/scala/wdl4s/values/WdlBoolean.scala
scottfrazer: I added that because I needed to use `WdlBoolean` in a `match` statement and withoutt his it wouldn't work.  http://danielwestheide.com/blog/2012/11/21/the-neophytes-guide-to-scala-part-1-extractors.html

src/main/java/wdl4s/parser/MemoryUnit.java
scottfrazer: Yeah, turns out the compiler doesn't like `1 << 40`.  It interprets it as `1 << 8` because ints are 32 bits.  `1L << 40` actually does produce the right number but it caused the tests to fail for a really weird reason.  I'll investigate.

src/main/scala/wdl4s/MemorySize.scala
scottfrazer: I'll see what I can do... the reason why I chose this approach is that I want the return type of `RuntimeAttributes.evaluate(...)` to be a `Map[String, Try[WdlValue]]`.  Since that function  is part of a public API, I wanted to avoid returning scalaz objects from it.  It seemed to make the most sense to to the conversion to `Try` here, though there are a few places I can do the conversion.

src/main/java/wdl4s/parser/MemoryUnit.java
scottfrazer: Turns out I'm wrong!  `1L << 40` does work.

scottfrazer:
Reviewers: @cjllanwarne and wheel chooses @kshakir

src/main/scala/wdl4s/RuntimeAttributes.scala
scottfrazer: Admittedly I didn't give much thought about how we ought to interpret integer values for memory.  Would you rather that an integer value for memory be interpreted as bytes?  I'm okay with that, just making sure we're on the same page.  It does make sense that `file_size()` would return the number of bytes a file is and not the number of gigabytes.

src/main/scala/wdl4s/MemorySize.scala
scottfrazer: https://github.com/broadinstitute/wdl/pull/27

src/main/scala/wdl4s/MemorySize.scala
geoffjentry: I would take that a step further and opine that a library should be throwing an exception and letting the consumer decide how to handle it. wdl4s doesn't currently do that in many places but that's an artifact of its heritage

scottfrazer:
@cjllanwarne your suggested test is good, though it needs to go in Cromwell because wdl4s currently lacks the ability to evaluate declarations and keep track of what has already been evaluated so the lookup function can be more dynamic.  You might argue that wdl4s should be able to do this and I think I agree, but currently it can't, so I added this test to Cromwell instead!

src/main/scala/wdl4s/RuntimeAttributes.scala
cjllanwarne: For me, yes, Bytes is my preference.

src/main/scala/wdl4s/MemorySize.scala
cjllanwarne: Thanks!

src/main/scala/wdl4s/values/WdlBoolean.scala
cjllanwarne: Every time I think I know scala...

Well TIL!

src/main/scala/wdl4s/MemorySize.scala
cjllanwarne: I didn't realise this was external-facing. I think Try would be my preference too in that case...

geoffjentry: The thing is, we should also try to get out of the habit of forcing Try on unsuspecting consumers as well, at least for new instances. There are places where Try is viewed just as suspiciously as other contentious constructs

scottfrazer: Weren't you were one of the people early on who was pushing really hard for `Try`s *instead* of throwing exceptions?  Don't you remember those arguments we had way back when where you argued that it's better to encapsulate the failure in the return type?  I'm sure @mcovarr remembers that too!

geoffjentry: Internally, yes. You should throw at the end of your system.

cjllanwarne: I actually think "chucking X on unsuspecting consumers" is exactly why Try is better than throwing an Exception.

I know I personally am much happier handling a Try than belatedly realising that some code I'm using throws exceptions and having to re-wrap it in a Try() myself

cjllanwarne:
:+1: from me.

src/main/scala/wdl4s/MemorySize.scala
scottfrazer: I agree with @cjllanwarne

src/main/scala/wdl4s/RuntimeAttributes.scala
```diff
@@ -1,76 +1,91 @@
...
-  private def validateMemoryUnit(unit: String): ValidationNel[String, MemorySize] = {
-    MemorySize.values find { _.suffixes.contains(unit) } match {
-      case Some(s) => s.successNel
-      case None => s"$unit is an invalid memory unit".failureNel
+  private def validateMemoryExpression(expression: WdlExpression, declarations: Seq[Declaration]): Try[WdlExpression] = {
+    // .get below because lookup functions should throw exceptions if they could not lookup the variable
+    def lookupType(n: String) = declarations.find(_.name == n).map(_.wdlType).get
+    expression.evaluateType(lookupType, new WdlStandardLibraryFunctionsType) match {
+      case Success(wdlType) if !Seq(WdlIntegerType, WdlStringType).contains(wdlType) =>
+        Failure(new SyntaxError("blah"))
```
scottfrazer: Best `SyntaxError` ever :)

src/main/scala/wdl4s/MemorySize.scala
geoffjentry: @cjllanwarne I don't care about this enough to continue the argument but my pushback on your point is this -> I think on the surface you're exactly right, however that Try in the signature provides a false sense of security as there's 0 guarantee it's not actually throwing some other exception on you. So then what do you do, wrap that response in its own Try?

src/main/scala/wdl4s/RuntimeAttributes.scala
```diff
@@ -1,76 +1,91 @@
...
     val asts = ast.findAsts(AstNodeName.Runtime)
     if (asts.size > 1) throw new UnsupportedOperationException("Only one runtime block may be defined per task")
-    val astList = asts.headOption map { _.getAttribute("map").asInstanceOf[AstList] }
-    val attrMap = astList map processRuntimeAttributes getOrElse Map.empty[String, Seq[String]]
-    attrMap.get("memory") flatMap { _.headOption } foreach validateMemory 
+    val kvPairAsts = asts.headOption.map(_.getAttribute("map").asInstanceOf[AstList].asScala.toVector.map(_.asInstanceOf[Ast]))
+    val runtimeAttributeMap = kvPairAsts match {
+      case Some(list) => list.map(ast => processRuntimeAttribute(ast, declarations)).toMap
+      case None => Map.empty[String, Try[WdlExpression]]
+    }
```
kshakir: `.toVector` plus `case Some(list)` are close to saying `val list: Vector[_]`.

kshakir:
Sorry for the delay. Only the one minor note, then LGTM

PR #13, opened 2016-02-17
[don't merge] proof of concept if-expressions

scottfrazer:
Currently the syntax is 

```
if (expression) true_expression else false_expression
```

Though we can also explore other syntax, like

```
expression ? true_expression : false_expression
true_expression if expression else false_expression
```

Some more complication of the ternary operator:

* For backends that need to know the outputs ahead of time, the ternary operator would need to be disallowed
```
task x {
  Boolean b
  command { ... }
  output {
    File out = if(b) "first_file" else "second_file"
  }
}
```

* We can't statically type-evaluate ternary expressions.  For example `if(b) "string" else 0` would either have to be disallowed or we have to accept that we can't know the type ahead of time.


src/main/scala/wdl4s/expression/TypeEvaluator.scala
```diff
@@ -91,6 +91,11 @@ case class TypeEvaluator(override val lookup: String => WdlType, override val fu
         case (_, Failure(ex)) => Failure(ex)
         case (_, _) => Failure(new WdlExpressionException(s"Can't index ${a.toPrettyString}"))
       }
+    case a: Ast if a.isTernaryOperator =>
+      (evaluate(a.getAttribute("true")), evaluate(a.getAttribute("false"))) match {
+        case (Success(l), Success(r)) if l == r => Success(l)
+        case (_, _) => Failure(new WdlExpressionException(s"True and false paths of ternary operator do not evaluate to the same type"))
```
cjllanwarne: Is this an issue if they can coerce to the same type?

cjllanwarne: Also, is this making assumptions about every other ternary operator?

src/main/scala/wdl4s/expression/ValueEvaluator.scala
```diff
@@ -124,6 +124,14 @@ case class ValueEvaluator(override val lookup: String => WdlValue, override val
         val name = a.getAttribute("name").sourceString
         val params = a.params map evaluate
         functions.getFunction(name)(params)
+      case a: Ast if a.isTernaryOperator =>
```
cjllanwarne: Again, not checking that the "ternary operator" in question is "if"

src/main/scala/wdl4s/expression/TypeEvaluator.scala
scottfrazer: Good catch... I'm not sure how to handle coercion, maybe we can't type-evaluate ternary operators statically if each branch is allowed to return a different type

src/main/scala/wdl4s/expression/ValueEvaluator.scala
scottfrazer: I guess it's just assumed that there's one ternary operator syntax

cjllanwarne:
@mcovarr is this really waiting for review?

mcovarr:
@cjllanwarne I really don't know what the story is with this

scottfrazer:
I'm going to close this for now.

PR #14, opened 2016-02-24
Fix coercion bug, allow for map function during command instantiation

scottfrazer:


kshakir:
:+1: Over to @Horneth for other review.

Horneth:
:+1:

PR #15, opened 2016-03-04
wdl4s piece of hashing, closes broadinstitute/cromwell#490

mcovarr:


src/main/scala/wdl4s/examples/ex5.scala
```diff
@@ -22,10 +22,9 @@ object ex5 {
 
     val ns = NamespaceWithWorkflow.load(wdl)
 
-    Seq(ns.resolve("wf.a"), ns.resolve("wf.b")) foreach { scope =>
-      scope match {
-        case Some(c: Call) => println(s"Call '${c.fullyQualifiedName}' prerequisites: ${c.prerequisiteScopes}")
-      }
+    Seq(ns.resolve("wf.a"), ns.resolve("wf.b")) foreach {
+      case Some(c: Call) => println(s"Call '${c.fullyQualifiedName}' prerequisites: ${c.prerequisiteScopes}")
+      case _ => throw new RuntimeException("Should never get here")
```
cjllanwarne: So... this catches a future regression or programmer error? Can you add some more details to help them correct their foolhardiness?

mcovarr: This is just moving code that was in Cromwell to wdl4s, you might want to look at that PR first.

src/main/scala/wdl4s/values/package.scala
```diff
@@ -0,0 +1,22 @@
+package wdl4s
+
+import org.apache.commons.codec.digest.DigestUtils
+
+
+package object values {
+
+  case class SymbolHash(value: String) extends Ordered[SymbolHash] {
```
cjllanwarne: extends AnyVal with Ordered\[SymbolHash\]?

cjllanwarne:
:+1:

src/main/scala/wdl4s/examples/ex5.scala
mcovarr: Chris and I discussed this in person, but in case anyone else sees this: I misremembered what I did here, this were some changes I made to muzzle a scalac warning (match not exhaustive) and an IntelliJ warning (use an anonymous PF).

src/main/scala/wdl4s/values/package.scala
mcovarr: apparently so

scottfrazer:
:+1:

PR #25, opened 2016-04-27
Add a workflow only static declarations evaluation method

Horneth:
`staticDeclarationsRecursive` evaluates all statics declarations, including the ones in a task. This won't work with the new evaluation design because we want those to be evaluated by the backend instead. This creates a `staticWorkflowDeclarationsRecursive` that only evaluates workflow declarations.

### DIFF
--- a/src/main/scala/wdl4s/WdlExpression.scala
+++ b/src/main/scala/wdl4s/WdlExpression.scala
@@ -109,7 +109,8 @@ object WdlExpression {
    */
   def standardLookupFunction(parameters: Map[String, WdlValue], declarations: Seq[Declaration], functions: WdlFunctions[WdlValue]): String => WdlValue = {
     def resolveParameter(name: String): Try[WdlValue] = parameters.get(name) match {
-      case Some(value) => Success(value)
+      case Some(value: WdlExpression) => Success(value.evaluate(lookup, functions)).getOrElse(Failure(new WdlExpressionException(s"Could not evaluate parameter: $value")))
+      case Some(value: WdlValue) => Success(value)
       case None => Failure(new WdlExpressionException(s"Could not resolve variable '$name' as an input parameter"))
     }
     def resolveDeclaration(lookup: ScopedLookupFunction)(name: String) = declarations.find(_.name == name) match {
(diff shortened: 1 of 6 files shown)


Horneth:
@mcovarr @scottfrazer

src/main/scala/wdl4s/WdlExpression.scala
```diff
@@ -109,7 +109,8 @@ object WdlExpression {
    */
   def standardLookupFunction(parameters: Map[String, WdlValue], declarations: Seq[Declaration], functions: WdlFunctions[WdlValue]): String => WdlValue = {
     def resolveParameter(name: String): Try[WdlValue] = parameters.get(name) match {
-      case Some(value) => Success(value)
+      case Some(value: WdlExpression) => Success(value.evaluate(lookup, functions)).getOrElse(Failure(new WdlExpressionException(s"Could not evaluate parameter: $value")))
```
mcovarr: `recoverWith` would probably be more straightforward

src/main/scala/wdl4s/WdlNamespace.scala
```diff
@@ -122,11 +122,18 @@ case class NamespaceWithWorkflow(importedAs: Option[String],
...
-    val attemptedEvaluations = declarationsByScope.flatMap(d => d.foldLeft(Map.empty[String, Try[WdlValue]])(evalDeclaration)).toMap
+    val attemptedEvaluations = scopedDeclarations.flatMap(d => d.foldLeft(Map.empty[String, Try[WdlValue]])(evalDeclaration)).toMap
     TryUtil.sequenceMap(attemptedEvaluations)
   }
 
+  def staticWorkflowDeclarationsRecursive(userInputs: WorkflowCoercedInputs, wdlFunctions: WdlStandardLibraryFunctions): Try[WorkflowCoercedInputs] = {
+    evaluateStaticDeclarations(userInputs, wdlFunctions, Seq(workflow.scopedDeclarations))
+  }
+
+  def staticDeclarationsRecursive(userInputs: WorkflowCoercedInputs, wdlFunctions: WdlStandardLibraryFunctions): Try[WorkflowCoercedInputs] = {
```
mcovarr: both of these would benefit from some Scaladoc to clarify the differences

mcovarr:
:+1:

Horneth:
see https://github.com/broadinstitute/wdl4s/pull/26

PR #44, opened 2016-11-02
~Do Not Merge~

ruchim:
Tests missing, needs rebase. Out for the purposes of early feedback/sanity checking.


src/main/scala/wdl4s/WdlNamespace.scala
```diff
@@ -143,86 +144,112 @@ case class WdlNamespaceWithWorkflow(importedAs: Option[String],
...
+      * For import statements without an 'as' clause, c# style include?
       */
-    val nonNamespacedImports = for {
+    val nonNamespacedImportsScopeAsts = for {
       imp <- imports if imp.namespaceName.isEmpty
-      source = importResolver(imp.uri)
-      ast = AstTools.getAst(source, imp.uri)
+      source = tryResolve(imp.uri)
+      if source.isDefined
+      ast = AstTools.getAst(source.get, imp.uri)
```
Horneth: `namespaces` above already creates a namespaces for all imports (with and without "as" clause).
Here imports without "as" clause are parsed a second time. It would be nice to find a way to avoid that.

src/main/scala/wdl4s/WdlNamespace.scala
```diff
@@ -143,86 +144,112 @@ case class WdlNamespaceWithWorkflow(importedAs: Option[String],
...
+    val mainWorkflowScopeAst = ast.getAttribute("body").astListAsVector.collect({ case a: Ast => a })
+
+    val topLevelScopeAsts = nonNamespacedImportsScopeAsts ++ mainWorkflowScopeAst
+
     /**
       * Map of Terminal -> WDL Source Code so the syntax error formatter can show line numbers
       */
-    val nonNamespacedImportsTerminalMap = nonNamespacedImports.flatMap({ case (a, s) => AstTools.terminalMap(a, s) }).toMap
-    val terminalMap = nonNamespacedImportsTerminalMap ++ AstTools.terminalMap(ast, source)
+    val terminalMap = AstTools.terminalMap(ast, source)
```
ruchim: Now that nonNamespacedImports are actually namespaced, I think this terminalMap contains everything.

src/main/scala/wdl4s/WdlNamespace.scala
```diff
@@ -143,86 +144,112 @@ case class WdlNamespaceWithWorkflow(importedAs: Option[String],
...
+
     val imports = for {
       importAst <- Option(ast).map(_.getAttribute("imports")).toSeq
       importStatement <- importAst.astListAsVector.map(Import(_))
     } yield importStatement
 
+    def tryResolve(str: String): Option[WdlSource] = {
+      importResolvers map { resolver =>
+        Try(resolver(str))
+      } collectFirst {
```
ruchim: Forgot to do this with my current PR, will make a note to fix this up!

PR #46, opened 2016-11-14
Imports & WdlNamespacing - re-review

ruchim:
Note: There's one new test case I introduced that's broken (Namespace name collision), but I wanted to start the review, and push that fix in soon.


src/main/scala/wdl4s/Import.scala
```diff
@@ -1,17 +1,21 @@
...
+
+    Import(uri, namespaceTerminal)
   }
 }
 
 // FIXME: I dislike dragging the AST along but it's necessary for "compile" time error syntax highlighting, argh
-case class Import(uri: String, namespaceTerminal: Option[Terminal]) {
-  val namespaceName: Option[String] = namespaceTerminal.map(_.sourceString)
+case class Import(uri: String, namespaceTerminal: Terminal) {
+  val namespaceName: String = File(namespaceTerminal.sourceString).nameWithoutExtension
```
Horneth: If the `namespaceTerminal` comes from the `as` clause you probably don't want/need to do `File(namespaceTerminal.sourceString).nameWithoutExtension`, although I imagine in most cases it wouldn't affect the alias string anyway.

src/main/scala/wdl4s/WdlNamespace.scala
```diff
@@ -403,7 +414,7 @@ object WdlNamespace {
             declarationName(decl.ast), wdlType, decl.wdlType
           )))
         case Success(wdlType) =>
-          expr.evaluate(NoLookup, NoFunctions) match {
+          expr.evaluate((s: String) => throw new Throwable("not implemented"), NoFunctions) match {
```
Horneth: Just curious, why this change ?

src/test/scala/wdl4s/SyntaxErrorSpec.scala
```diff
@@ -124,33 +125,8 @@ class SyntaxErrorSpec extends FlatSpec with Matchers {
       """.stripMargin
   }
 
-  case object TaskAndNamespaceNameCollision extends ErrorWdl {
```
Horneth: So you decided to remove this one in the end ?

src/test/scala/wdl4s/SyntaxErrorSpec.scala
```diff
@@ -610,9 +587,8 @@ class SyntaxErrorSpec extends FlatSpec with Matchers {
     CallReferencesBadInput,
     CallReferencesBadTask,
     MultipleWorkflows,
-    TaskAndNamespaceNameCollision,
     WorkflowAndNamespaceNameCollision,
-    TwoTasksHaveTheSameName,
+    //NamespaceNameCollision,
```
Horneth: Should this be uncommented ?

src/main/scala/wdl4s/WdlNamespace.scala
```diff
@@ -143,76 +147,82 @@ case class WdlNamespaceWithWorkflow(importedAs: Option[String],
...
-      scopeAst <- ast.getAttribute("body").astListAsVector.collect({ case a: Ast => a })
-    } yield ast -> source
-
-    val topLevelScopeAsts = for {
-      namespaceAst <- nonNamespacedImports.map(_._1) :+ ast
-      scopeAst <- namespaceAst.getAttribute("body").astListAsVector.collect({ case a: Ast => a })
-    } yield scopeAst
+    val namespaces: Seq[WdlNamespace] = for {
+      imp <- imports
+      source = tryResolve(imp.uri, importResolvers, List.empty)
```
cjllanwarne: Since `tryResolve` will fail if you try to pass in 0 resolvers (because the `Nil` case requires at least one error), make the `importResolvers` parameter a `NonEmptyList`

src/main/scala/wdl4s/WdlNamespace.scala
```diff
@@ -143,76 +147,82 @@ case class WdlNamespaceWithWorkflow(importedAs: Option[String],
...
-    /**
-      * For import statements with an 'as' clause, this translates to a sub-namespace
-      */
-    val namespaces: Seq[WdlNamespace] = for {
-      imp <- imports if imp.namespaceName.isDefined
-      source = importResolver(imp.uri)
-      if source.length > 0
-    } yield WdlNamespace.load(source, imp.uri, importResolver, imp.namespaceName)
+
+    def tryResolve(str: String, remainingResolvers: Seq[ImportResolver], errors: List[Throwable]): Option[WdlSource] = {
```
cjllanwarne: ToL: Add comment: `Make sure remainingResolvers has at least one entry the first time this is called`

cjllanwarne: This might as well be returning `WdlSource` since we never return `None`

src/main/scala/wdl4s/WdlNamespace.scala
```diff
@@ -143,76 +147,82 @@ case class WdlNamespaceWithWorkflow(importedAs: Option[String],
...
+    load(wdlSource, resource, Seq(fileResolver), None)
   }
 
   def load(wdlSource: WdlSource, resource: String, importResolver: ImportResolver): WdlNamespace = {
-    load(wdlSource, resource, importResolver, None)
+    load(wdlSource, resource, Seq(importResolver), None)
   }
 
-  private def load(wdlSource: WdlSource, resource: String, importResolver: ImportResolver, importedAs: Option[String]): WdlNamespace = {
+  private def load(wdlSource: WdlSource, resource: String, importResolver: Seq[ImportResolver], importedAs: Option[String]): WdlNamespace = {
```
cjllanwarne: `importedAs` might as well be a `String`. We should always have something to import it as now

src/main/scala/wdl4s/WdlNamespace.scala
```diff
@@ -321,10 +331,11 @@ object WdlNamespace {
 
     def scopeNameAndTerminal(scope: Scope): (String, Terminal) = {
       scope match {
-        case ns: WdlNamespace => ("Namespace", imports.find(_.namespaceName == ns.importedAs).get.namespaceTerminal.get)
+        case ns: WdlNamespace => ("Namespace", imports.find(_.namespaceName == ns.importedAs.get).get.namespaceTerminal)
```
cjllanwarne: ToL: This'll look a lot nicer if these aren't `Option`s (as suggested above)

src/main/scala/wdl4s/formatter/SyntaxFormatter.scala
```diff
@@ -79,8 +79,8 @@ class SyntaxFormatter(highlighter: SyntaxHighlighter = NullSyntaxHighlighter) {
   }
 
   private def formatImport(imp: Import): String = {
-    val namespace = imp.namespaceTerminal.map(ns => s" as ${ns.sourceString}").getOrElse("")
-    s"${highlighter.keyword("import")} '${imp.uri}'$namespace"
+    val namespace = imp.namespaceName
+    s"${highlighter.keyword("import")} '${imp.uri}' as $namespace"
```
cjllanwarne: ToL: Hmmm, this is used when highlighting source code? I think it's potentially a little confusing if we're return edited source code instead?

src/test/cases/three_step_import/ancestry.expectations
```diff
@@ -1,10 +1,12 @@
 {
   "cgrep": [""],
-  "cgrep.count": ["cgrep", ""],
-  "cgrep.in_file": ["cgrep", ""],
-  "cgrep.pattern": ["cgrep", ""],
+  "cgrep.cgrep": ["cgrep", ""],
```
cjllanwarne: Scott's changes? I'll ignore all these

src/test/scala/wdl4s/SyntaxErrorSpec.scala
```diff
@@ -173,28 +149,29 @@ class SyntaxErrorSpec extends FlatSpec with Matchers {
       """.stripMargin
   }
 
-  case object TwoTasksHaveTheSameName extends ErrorWdl {
-    val testString = "detect when two tasks have the same name"
+  //RUCHI: Needs fixing, the syntax formatter should list both import statements, not
```
cjllanwarne: TODONE?

src/test/scala/wdl4s/SyntaxErrorSpec.scala
```diff
@@ -173,28 +149,29 @@ class SyntaxErrorSpec extends FlatSpec with Matchers {
...
-        |Task defined here (line 2, col 6):
+        |Namespace defined here (line 1, col 16):
         |
-        |task ps {
-        |     ^
+        |import "ps" as ps
+        |               ^
         |
-        |Task statement defined here (line 2, col 6):
+        |Namespace defined here (line 1, col 16):
```
cjllanwarne: Presumably these two `(line 1, col 16)` references should be different?

src/test/scala/wdl4s/SyntaxHighlightSpec.scala
```diff
@@ -163,8 +182,7 @@ class SyntaxHighlightSpec extends Matchers with WordSpecLike {
         |  call t as u {
         |    input: f="abc", p=p
         |  }
-        |}""".stripMargin,
-      importResolver = (s:String) => ""
+        |}""".stripMargin, resolver _
```
cjllanwarne: ToL: I think you can change resolver to remove this `_`
```
def resolver: String => WdlSource = (importUri: String) => {
  ...
}
```

cjllanwarne: ToL: bring the closing `)` up?

src/test/scala/wdl4s/ThreeStepImportSpec.scala
```diff
@@ -63,14 +63,14 @@ class ThreeStepImportSpec extends FlatSpec with Matchers {
...
+    namespace.tasks.size shouldEqual 0
   }
 
   it should "Have 0 imported WdlNamespaces" in {
-    namespace.namespaces.size shouldEqual 0
+    namespace.namespaces.size shouldEqual 3
   }
   it should "Have tasks with the names 'ps', 'cgrep' and 'wc'" in {
-    namespace.tasks map {_.name} shouldEqual Seq("ps", "cgrep", "wc")
+    namespace.namespaces flatMap {_.tasks} map {_.name} shouldEqual Seq("ps", "cgrep", "wc")
```
cjllanwarne: Misleading test name: `should "Have namespaces with the names ..."`

src/test/scala/wdl4s/WorkflowImportsSpec.scala
```diff
@@ -0,0 +1,261 @@
...
+    |
+    |}
+    |""".stripMargin
+
+  val wdlWithImports = imports + primaryWorkflow
+
+  val namespace = WdlNamespaceWithWorkflow.load(wdlWithImports, wdlDirectory)
+
+  "WDL file with imports" should "Have 0 tasks (3 tasks are in separate namespace)" in {
+    namespace.tasks.size shouldEqual 1
```
cjllanwarne: `should "Have 1 tasks...`?

src/main/scala/wdl4s/WdlNamespace.scala
ruchim: @cjllanwarne Upon investigation, this optional string seem to influence how the typeEvaluator and valueEvaluator functions work -- they depend on being able to filter out WdlNamespaces with an undefined value for importedAs. So although all imports now have a namespaceName, the primary workflow still doesn't, so it truly is optional for WdlNamespace objects. This could be fixed with some restructuring...thoughts?

src/main/scala/wdl4s/WdlNamespace.scala
ruchim: As mentioned above, by default we pass in atleast 1 resolver. If this method `tryResolve` is called from elsewhere, now it throws a proper message saying no import resolvers were provided.

ruchim:
@Horneth @cjllanwarne I'm waiting for the official thumbs, please let me know if I missed any of your requests.

Horneth:
👍

src/main/scala/wdl4s/WdlNamespace.scala
```diff
@@ -186,17 +168,20 @@ object WdlNamespace {
       importStatement <- importAst.astListAsVector.map(Import(_))
     } yield importStatement
 
-
-    def tryResolve(str: String, remainingResolvers: Seq[ImportResolver], errors: List[Throwable]): Option[WdlSource] = {
+    //TODO: Make sure remainingResolvers has at least one entry the first time this is called
```
cjllanwarne: Since you already handle the `case Nil` below, I think this is TODONE.

cjllanwarne:
👍

PR #47, opened 2016-11-15
Sub Workflows

Horneth:


Horneth:
Looks like all the *.html are scala docs updates due to some class renaming

geoffjentry:
@Horneth Along with the rest of the side things which need updating can you make a PR against the spec in the wdl repo to reflect the changes? Also I think that'd be helpful in general when trying to digest the PR once everything is ready

Horneth:
@geoffjentry Yep. I'm also not sure where / how much doc we want to generate for sub workflows in general. Is the Cromwell README still a good place (since it's currently being rewritten by Kate) ?

I think there are 2 different pieces : the WDL spec, and how Cromwell specifically executes sub workflows (or at least the user facing implications).

geoffjentry:
@Horneth For this I'm thinking specifically the WDL spec - it came to mind when I was looking through the grammar file and such.

cjllanwarne:
👍

mcovarr:
beyond the noise factor, aren't the HTML changes to earlier wdl4s versions incorrect since those were in fact the names of those classes in those releases?

Horneth:
Yes - I reverted them but forgot to push

mcovarr:
:+1: hooray

PR #49, opened 2016-11-16
Optional types as 1st class citizens in WDL

cjllanwarne:
Significantly updated since the last two-👍 PR. Ready for a re-review.


src/main/scala/wdl4s/expression/WdlStandardLibraryFunctions.scala
```diff
@@ -138,6 +138,20 @@ trait WdlStandardLibraryFunctions extends WdlFunctions[WdlValue] {
     } yield WdlString(pattern.r.replaceAllIn(str, replace))
   }
 
+  def select_first(params: Seq[Try[WdlValue]]): Try[WdlValue] = extractSingleArgument(params) flatMap {
+    case WdlArray(WdlArrayType(WdlOptionalType(_)), arrayValue) =>
+      (arrayValue collect { case WdlOptionalValue(_, Some(wdlValue)) => wdlValue }).headOption
```
Horneth: `collectFirst` would get rid of the `headOption` I think

src/main/scala/wdl4s/expression/WdlStandardLibraryFunctions.scala
```diff
@@ -262,6 +276,12 @@ class WdlStandardLibraryFunctionsType extends WdlFunctions[WdlType] {
     case Success(t @ WdlArrayType(WdlArrayType(wdlType))) :: Nil => Success(t)
     case _ => Failure(new Exception(s"Unexpected transpose target: $params"))
   }
+  def select_first(params: Seq[Try[WdlType]]): Try[WdlType] = extractSingleArgument(params) flatMap {
+    case WdlArrayType(WdlOptionalType(innerType)) => Success(innerType)
```
Horneth: Can we add a case for other types too with a failure like in the value implementation ?
It's still better than having a hard to track scala "match error".

src/main/scala/wdl4s/expression/WdlStandardLibraryFunctions.scala
```diff
@@ -262,6 +276,12 @@ class WdlStandardLibraryFunctionsType extends WdlFunctions[WdlType] {
...
+  }
+  def select_all(params: Seq[Try[WdlType]]): Try[WdlType] = extractSingleArgument(params) flatMap {
+    case WdlArrayType(WdlOptionalType(innerType)) => Success(WdlArrayType(innerType))
```
Horneth: Same as above

src/main/scala/wdl4s/types/WdlOptionalType.scala
```diff
@@ -0,0 +1,36 @@
...
+    * construct `WdlBoolean`s for inputs of supported types and contents.  Values for which
+    * the partial function is not defined are assumed to not be convertible to the target type.
+    */
+  override protected def coercion: PartialFunction[Any, WdlValue] = {
+
+    // It's safe to box up values implicitly:
+    case wdlValue: WdlValue if memberType.equals(wdlValue.wdlType) => WdlOptionalValue(wdlValue)
+    case coerceable: Any if memberType.coercionDefined(coerceable) => WdlOptionalValue(memberType.coerceRawValue(coerceable).get)
+
+
```
Horneth: maybe 1 `\n` is enough ? :p

Horneth:
👍

src/main/scala/wdl4s/values/WdlOptionalValue.scala
```diff
@@ -0,0 +1,18 @@
+package wdl4s.values
+
+import wdl4s.types.{WdlOptionalType, WdlType}
+
+case class WdlOptionalValue(innerType: WdlType, value: Option[WdlValue]) extends WdlValue {
+  override val wdlType = WdlOptionalType(innerType)
+  override val toWdlString = value map { _.toWdlString } getOrElse "null"
```
kshakir: ToL: 

I'm curious the effective behavior of a round trip. I believe this works mainly the same was as `String`, in that a round trip of a None would end up being a Some("null"), similar to:

```scala
scala> val s: String = null
s: String = null

scala> val v = String.valueOf(s)
v: String = null

scala> v.length()
res0: Int = 4

scala> 
```

Where does "null" show up? If one (effectively) does:

```
Optional[String] a

command {
  echo "hello ${a} world"
}
```

Is the output:
a) `hello  world`
b) `hello null world`
c) This isn't a valid use case, as a task with this PR's type of optional should never run its command.

kshakir:
LGTM 👍 , with a ToL question above.

src/main/scala/wdl4s/expression/WdlStandardLibraryFunctions.scala
```diff
@@ -138,6 +138,20 @@ trait WdlStandardLibraryFunctions extends WdlFunctions[WdlValue] {
...
+      (arrayValue collect { case WdlOptionalValue(_, Some(wdlValue)) => wdlValue }).headOption
+        .map(Success(_))
+        .getOrElse(Failure(new IllegalArgumentException("Select_first failed. All provided values were empty.")))
+    case other => Failure(new IllegalArgumentException("Select_first must take an array of optional values but got: " + other.toWdlString))
+  }
+
+  def select_all(params: Seq[Try[WdlValue]]): Try[WdlArray] = extractSingleArgument(params) flatMap {
+    case WdlArray(WdlArrayType(WdlOptionalType(memberType)), arrayValue) =>
+      Success(WdlArray(WdlArrayType(memberType), arrayValue collect { case WdlOptionalValue(_, Some(wdlValue)) => wdlValue }))
+    case other => Failure(new IllegalArgumentException("Select_all must take an array of optional values but got: " + other.toWdlString))
```
mcovarr: probably shouldn't capitalize `select_all`

src/main/scala/wdl4s/expression/WdlStandardLibraryFunctions.scala
```diff
@@ -138,6 +138,20 @@ trait WdlStandardLibraryFunctions extends WdlFunctions[WdlValue] {
     } yield WdlString(pattern.r.replaceAllIn(str, replace))
   }
 
+  def select_first(params: Seq[Try[WdlValue]]): Try[WdlValue] = extractSingleArgument(params) flatMap {
+    case WdlArray(WdlArrayType(WdlOptionalType(_)), arrayValue) =>
+      (arrayValue collect { case WdlOptionalValue(_, Some(wdlValue)) => wdlValue }).headOption
+        .map(Success(_))
+        .getOrElse(Failure(new IllegalArgumentException("Select_first failed. All provided values were empty.")))
+    case other => Failure(new IllegalArgumentException("Select_first must take an array of optional values but got: " + other.toWdlString))
```
mcovarr: uncap `select_first`

src/main/scala/wdl4s/values/WdlOptionalValue.scala
cjllanwarne: To answer your first question, there is no possible round trip. You can go `value => value?` but you cannot go `value? => value` without the select_x functions.

The second question, it would give you `hello  world` since the string interpolation is a separate function from `toWdlString` (just like `"echo ${array}"` doesn't print out the WDL array literal like `[1,2,3]`). `null` might be the WDL literal that evaluates to None although that's not-yet-implemented.

mcovarr:
:+1:

src/main/scala/wdl4s/expression/WdlStandardLibraryFunctions.scala
cjllanwarne: done!

mcovarr:
compilation failure??

cjllanwarne:
@mcovarr but only on pr, not push...?

PR #50, opened 2016-11-16
Fix Pair validation bug Closes #48

### LINKED ISSUE #48: Call input validation expects LHS of MemberAccess in the input mappings to be Calls, but now they can be scatter items too if the item is a pair.

Horneth:


cjllanwarne:
Oops! Thanks for fixing this! 👍

kshakir:
LGTM 👍

PR #52, opened 2016-11-22
Declaration evaluation

Horneth:
In order to allow for workflow declarations referencing call outputs, the WDL validation needs to be a little bit more flexible and not fail right away if it can't evaluate a workflow declaration at submission time.
- The first commit modifies `staticDeclarationsRecursive` so it doesn't reject the namespace if it can't evaluate a declaration because it couldn't find a call output.
- The second commit makes sure the underlying exception is bubbled up from the lookup function and not a wrapped one.
- The third commit is just a rename for clarity and has nothing to do with the rest.

### DIFF
--- a/src/main/scala/wdl4s/Scope.scala
+++ b/src/main/scala/wdl4s/Scope.scala
@@ -134,15 +134,15 @@ trait Scope {
     * @return closest common ancestor
     */
   def closestCommonAncestor(other: Scope): Option[Scope] = {
-    val otherAncestry = other.ancestry
-    ancestry find { otherAncestry.contains(_) }
+    val otherAncestry = other.ancestrySafe
+    ancestrySafe find { otherAncestry.contains(_) }
   }
 
   /**
     * Performs scope resolution starting from this scope and walking up the lexical hierarchy
     * until it finds a GraphNode with the `name` as its unqualifiedName
     */
-  def resolveVariable(name: String, relativeTo: Scope = this): Option[Scope with GraphNode] = {
+  def resolveVariable(name: String, relativeTo: Scope = this): Option[GraphNode] = {
     val siblingScopes = if (children.contains(relativeTo))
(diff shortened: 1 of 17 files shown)


src/main/scala/wdl4s/Scope.scala
```diff
@@ -183,59 +183,65 @@ trait Scope {
...
       }
     }
 
     def lookup(name: String): WdlValue = {
-      val scopeResolvedValue = resolveVariable(name, relativeTo) flatMap {
+      val scopeResolvedValue = resolveVariable(name, relativeTo) map {
         // First check if the variable has been provided in the known values
-        case scope if knownInputs.contains(scope.fullyQualifiedName) => knownInputs.get(scope.fullyQualifiedName)
+        case scope if knownInputs.contains(scope.fullyQualifiedName) => 
+          knownInputs.get(scope.fullyQualifiedName) map Success.apply getOrElse {
```
geoffjentry: On all of these `map Success.apply` - you're converting from `Option` to `Try`, right? If so, what about using a pattern match on these.

Also I could have sworn we had an implicit class somewhere to do this sort of thing but it's probably not in wdl4s

geoffjentry:
👍

src/main/scala/wdl4s/Scope.scala
Horneth: I don't remember seeing the implicit but I'll change it to a pattern match 👍

mcovarr: For a hypothetical implicit, one can definitely go Try => Option, but how would the None handling work in the Option => Try case?

mcovarr:
:+1:

src/main/scala/wdl4s/Scope.scala
geoffjentry: I'm not suggesting one be created, just that I thought we had one somewhere. As to how it would work, dunno - my assumption would be some stupid exception.

PR #53, opened 2016-11-26
Allow for meta and parameter_meta blocks in the workflow section

geoffjentry:
Closes WDL issue #67. PRs for related repos (WDL, wdltool & cromwell) to follow after this is merged


src/main/scala/wdl4s/AstTools.scala
```diff
@@ -286,6 +286,26 @@ object AstTools {
 
   def terminalMap(ast: Ast, source: WdlSource) = (findTerminals(ast) map {(_, source)}).toMap
 
+  def wdlSectionToStringMap(ast: Ast, node: String, wdlSyntaxErrorFormatter: WdlSyntaxErrorFormatter): Map[String, String] = {
```
geoffjentry: this was refactored nearly verbatim from `Task`

mcovarr:
:+1:

src/main/scala/wdl4s/AstTools.scala
```diff
@@ -286,6 +286,26 @@ object AstTools {
...
+    ast.findAsts(node) match {
+      case a if a.isEmpty => Map.empty[String, String]
+      case a if a.size == 1 =>
+        // Yes, even 'meta {}' and 'parameter_meta {}' sections have RuntimeAttribute ASTs.
+        // In hindsight, this was a poor name for the AST.
```
geoffjentry: Touching on our convo this morning, probably not hard but I'd be real leery of hosing things up. 

The name definitely is confusing, if it weren't for that comment I'd not have realized I needed this bit.

mcovarr: Are you concerned the existing tests wouldn't notice if you did in fact hose this up?

Horneth:
👍  with maybe a small test to make sure that the pairs make it to the `meta` and `parameter_meta` attributes in the workflow once the namespace is loaded ?

geoffjentry:
@Horneth I'll double check but I don't think the one test I modified would pass if that weren't the case

src/main/scala/wdl4s/AstTools.scala
cjllanwarne: My ToL 2c, this seems possibly worth hosing-up (but then obviously also fixing-up), if possible?

src/main/scala/wdl4s/AstTools.scala
cjllanwarne: Refactored of copy/pasted? Is it DRYable?

geoffjentry: The whole point of moving it here was to not copy paste

src/main/scala/wdl4s/AstTools.scala
cjllanwarne: Looks like the offender is:
```
     $kv = :identifier :colon $e -> RuntimeAttribute(key=$0, value=$2)
```
... which is causing all the key/value pairs to create RuntimeAttribute ASTs. You could rename it something more generic like `KeyValuePair`, and update all wdl4s builders to search for that instead of `RuntimeAttribute`. Hopefully...

geoffjentry: Discussed earlier in slack with @mcovarr - my concern was that the original author of that comment is someone I would have expected to fix that if it wasn't a big deal. I can see reasons why he wouldn't have that don't fit that model however.

cjllanwarne: Yeah. My refined ToL preference would actually be, disallow expressions in meta blocks... but (shrug)

geoffjentry: I believe @cjllanwarne is correct. As the underlying AST structure is tangential to this change and because I'd like to get this change in the release (and I don't have faith I'd manage to get both squared away) I filed [wdl issue 73](https://github.com/broadinstitute/wdl/issues/73).

I can't imagine there's a reason to *not* change the grammar for that (in fact it would more closely adhere to my reading of the spec) but we shouldn't change functionality like that w/o PO discussion, etc.

geoffjentry:
@Horneth Double checked - unless I misunderstood you that's explicitly checked by the one key test as that content is coming from the workflow/task objects

PR #59, opened 2016-12-07
Wdl4s changes to support the graph printout in wdltool

cjllanwarne:
Most of these changes are easiest to see in the changes to the expectations files. So, if you want a good place to start reviewing, those files would be a good start.


src/main/scala/wdl4s/GraphNode.scala
```diff
@@ -1,6 +1,32 @@
...
+
+  final lazy val upstream: Set[GraphNode] = {
+    val referencedNodes = for {
+      variable <- upstreamReferences
+      node <- resolveVariable(variable.sourceString)
+      if node.fullyQualifiedNameWithIndexScopes != fullyQualifiedNameWithIndexScopes
+    } yield node
+
+    val closestScopedAncestor = ancestry.collectFirst({
+      case s: Scope with GraphNode => s
```
Horneth: `GraphNode` extends `Scope` so this can just be `GraphNode`

Horneth:
👍

src/test/cases/coercion_not_defined/ancestry.expectations
```diff
@@ -0,0 +1,15 @@
+{
+  "summary": [""],
```
cjllanwarne: This is weird. Looks like the `""` in all this ancestry represents the nameless default namespace.

Maybe we could tidy this later... but since this wasn't a change caused by this PR I'll let it slide for now.

src/main/scala/wdl4s/GraphNode.scala
```diff
@@ -1,6 +1,69 @@
...
+
+  final lazy val upstreamAncestry: Set[GraphNode] = GraphNode.calculateUpstreamAncestry(Set.empty, this)
+
+  def referencedNodes: Iterable[GraphNode]
+
+  final lazy val upstream: Set[GraphNode] = {
+
+    if (this.fullyQualifiedName.equals("test1.summary")) {
+      val x = 5
+    }
```
Horneth: That looks like it shouldn't be there

src/main/scala/wdl4s/GraphNode.scala
```diff
@@ -1,6 +1,69 @@
...
+    } yield node
+  }
 }
+
+object GraphNode {
+  private def calculateUpstreamAncestry(currentSet: Set[GraphNode], graphNode: GraphNode): Set[GraphNode] = {
+
+    if (graphNode.fullyQualifiedName.equals("w.$scatter_0")) {
+      val x = 5
+    }
```
Horneth: nix

src/main/scala/wdl4s/GraphNode.scala
```diff
@@ -1,6 +1,69 @@
...
+      node <- namespace.descendants.collect({
+        case n: GraphNode if n.fullyQualifiedName != fullyQualifiedName => n
+      })
+      if node.upstream.contains(this)
+    } yield node
+  }
 }
+
+object GraphNode {
+  private def calculateUpstreamAncestry(currentSet: Set[GraphNode], graphNode: GraphNode): Set[GraphNode] = {
```
Horneth: an explanation on what this is exactly calculating would be nice.

src/main/scala/wdl4s/GraphNode.scala
```diff
@@ -1,6 +1,69 @@
...
+
+  // If we have variable reference to other graph nodes, then they are upstream from us.
+  override final def referencedNodes = for {
+      variable <- upstreamReferences
+      node <- resolveVariable(variable.sourceString)
+      if node.fullyQualifiedNameWithIndexScopes != fullyQualifiedNameWithIndexScopes
+    } yield node
+}
+
+trait GraphNodeWithInputs extends GraphNode {
```
Horneth: Doesn't look like this is mixed anywhere else than in `Call`. Do we still need a trait ?

cjllanwarne:
This PR will return...

PR #60, opened 2016-12-09
Operations for optional values

Horneth:
Closes https://github.com/broadinstitute/cromwell/issues/1765

### DIFF
--- a/src/main/scala/wdl4s/values/WdlString.scala
+++ b/src/main/scala/wdl4s/values/WdlString.scala
@@ -12,21 +12,29 @@ case class WdlString(value: String) extends WdlPrimitive {
     case r: WdlInteger => Success(WdlString(value + r.value))
     case r: WdlFloat => Success(WdlString(value + r.value))
     case r: WdlFile => Success(WdlString(value + r.value))
+    case WdlOptionalValue(_, Some(r)) => add(r)
+    case r: WdlOptionalValue => emptyValue(r)
     case _ => invalid(s"$value + $rhs")
   }
 
   override def equals(rhs: WdlValue): Try[WdlBoolean] = rhs match {
     case r: WdlString => Success(WdlBoolean(value == r.value))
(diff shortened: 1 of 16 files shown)


cjllanwarne:
Should we double check the semantics of this?

E.g. as it currently is, we fail the workflow if we try to do (e.g.) 
```
Int? x  // unsupplied
Int y = x + 5
```

But I could imagine what we really want to do is map the Options, e.g.:
```
Int? x
Int? y = x + 5
```

Horneth:
Yes we should definitely talk about it, I should have put the request for feedback label.

src/main/scala/wdl4s/values/WdlString.scala
```diff
@@ -12,21 +12,29 @@ case class WdlString(value: String) extends WdlPrimitive {
...
     case r: WdlString => Success(WdlBoolean(value == r.value))
+    case WdlOptionalValue(_, Some(r)) => equals(r)
+    case r: WdlOptionalValue => emptyValue(r)
     case _ => invalid(s"$value == $rhs")
   }
 
   override def lessThan(rhs: WdlValue): Try[WdlBoolean] = rhs match {
     case r: WdlString => Success(WdlBoolean(value < r.value))
+    case WdlOptionalValue(_, Some(r)) => lessThan(r)
+    case r: WdlOptionalValue => emptyValue(r)
```
cjllanwarne: I'm sure there's some clever way to DRY all these cases, e.g.
```
def operatorIfSet[A, B](wdlOptional: WdlOptionalValue, ifSet: A=> B) = wdlOptional match {
  case WdlOptionalValue(_. Some(a)) => ifSet(a)
  else emptyValue(wdlOptional)
}
```
and then in this situation:
```
override def lessThan[...] {
  [...]
  case o: WdlOptionalValue => operatorIfSet(o, lessThan _)
```

cjllanwarne:
Otherwise, LGTM 👍

mcovarr:
:+1:

PR #65, opened 2017-01-06
Add length() engine function for Array lengths.  Closes broadinstitute/cromwell#1604

mcovarr:


src/main/scala/wdl4s/expression/WdlStandardLibraryFunctions.scala
```diff
@@ -100,6 +100,13 @@ trait WdlStandardLibraryFunctions extends WdlFunctions[WdlValue] {
     extractExactlyOneArg.flatMap(validateAndExpand).flatMap(transpose)
   }
 
+  def length(params: Seq[Try[WdlValue]]): Try[WdlInteger] = {
+    params.toList match {
+      case Success(WdlArray(_, arrayValues)) :: Nil => Success(WdlInteger(arrayValues.length))
+      case _ => Failure(new UnsupportedOperationException(s"length() expects one parameter of type Array but got ${params.length} instead."))
```
cjllanwarne: I'd change this to:
```
... but got [${params.map(_.wdlType.toWdlString.mkString(", "))}] instead.
```

cjllanwarne: Well obviously nicer that *that* but you get the idea...

src/main/scala/wdl4s/expression/WdlStandardLibraryFunctions.scala
```diff
@@ -276,6 +283,7 @@ class WdlStandardLibraryFunctionsType extends WdlFunctions[WdlType] {
   def write_json(params: Seq[Try[WdlType]]): Try[WdlType] = Success(WdlFileType)
   def glob(params: Seq[Try[WdlType]]): Try[WdlType] = Success(WdlArrayType(WdlFileType))
   def size(params: Seq[Try[WdlType]]): Try[WdlType] = Success(WdlFloatType)
+  def length(params: Seq[Try[WdlType]]): Try[WdlType] = Success(WdlIntegerType)
```
cjllanwarne: Worth checking that the incoming type is an array?

kshakir:
👍  with ToL reservations: 

1) The multitude of ways checking if `Seq[Try[WdlValue]]` contains exactly one element in WdlStandardLibraryFunctions.scala is excessive.
- `params.size match { case 1 => `
- `params.toList match { case Success(…) :: Nil => `
- `extractSingleArgument("method_name_here", params) map { case Success(…) => `
- and maybe more?

2) Also concur that length's tests should be in wdl4s, not cromwell. (Eventually, with transpose's specs?)

geoffjentry:
@kshakir speaks the truth

src/main/scala/wdl4s/expression/WdlStandardLibraryFunctions.scala
```diff
@@ -1,7 +1,7 @@
 package wdl4s.expression
 
 import wdl4s.types._
-import wdl4s.values._
+import wdl4s.values.{WdlArray, WdlInteger, WdlValue, _}
```
mcovarr: This is IntelliJ mischief, not a conscious choice on my part.  I'll put it back if IntelliJ allows me.

src/main/scala/wdl4s/expression/WdlStandardLibraryFunctions.scala
mcovarr: None of the other functions are checking any inbound types...

cjllanwarne: Some of them are. But on investigation those are the ones which give different result types depending on the input so this might be ok...

cjllanwarne:
Oops I forgot to 👍 this

src/main/scala/wdl4s/expression/WdlStandardLibraryFunctions.scala
mcovarr: oh whoops I missed those

PR #67, opened 2017-01-17
Optional evaluations and better errors. Closes broadinstitute/cromwell#1830

cjllanwarne:


src/main/scala/wdl4s/values/WdlValue.scala
```diff
@@ -11,10 +11,10 @@ import scala.util.{Failure, Try}
 trait WdlValue {
   val wdlType: WdlType
   def invalid(operation: String) = Failure(new WdlExpressionException(s"Cannot perform operation: $operation"))
-  def emptyValue(value: WdlOptionalValue) = Failure(VariableNotFoundException(value.toString))
-  def evaluateIfDefined[A <: WdlValue](optionalValue: WdlOptionalValue, operation: WdlValue => Try[A]): Try[A] = optionalValue match {
+  def emptyValueFailure(operationName: String) = Failure(new UnsupportedOperationException(s"Sorry! Operation $operationName is not supported on empty optional values. You might resolve this using select_first(optional, default) to guarantee that you have a filled value."))
```
cjllanwarne: Pinging @katevoss for critique!

cjllanwarne: Maybe should be `select_first([optional, default])`

mcovarr: filled?  like a doughnut?

mcovarr:
:+1: monads are like burritos, options are like doughnuts.

cjllanwarne:
@katevoss that `+` is from a test case which tries something invalid with `+`. Whatever `operationName` is  gets filled in into the error message as appropriate.

Horneth:
👍

cjllanwarne:
Will return later!

PR #72, opened 2017-01-23
Workflow outputs fix

Horneth:
Do not replace . with _ when expanding workflow outputs in old notation to new notation.
Throw an exception when the namespace is built if a workflow is called as a sub workflow and has old syntax.

### DIFF
--- a/src/main/scala/wdl4s/WdlNamespace.scala
+++ b/src/main/scala/wdl4s/WdlNamespace.scala
@@ -173,8 +173,8 @@ object WdlNamespace {
     load(wdlSource, resource.getOrElse("string"), importResolver.getOrElse(Seq(fileResolver)), None)
   }
 
-  private def load(wdlSource: WdlSource, resource: String, importResolver: Seq[ImportResolver], importedAs: Option[String]): WdlNamespace = {
-    WdlNamespace(AstTools.getAst(wdlSource, resource), wdlSource, importResolver, importedAs, root = true)
+  private def load(wdlSource: WdlSource, resource: String, importResolver: Seq[ImportResolver], importedAs: Option[String], root: Boolean = true): WdlNamespace = {
+    WdlNamespace(AstTools.getAst(wdlSource, resource), wdlSource, importResolver, importedAs, root = root)
   }
 
 
@@ -206,7 +206,7 @@ object WdlNamespace {
     val namespaces: Seq[WdlNamespace] = for {
       imp <- imports
       source = tryResolve(imp.uri, importResolvers, List.empty)
-    } yield WdlNamespace.load(source, imp.uri, importResolvers, Option(imp.namespaceName))
+    } yield WdlNamespace.load(source, imp.uri, importResolvers, Option(imp.namespaceName), root = false)
 
     /**
       * Map of Terminal -> WDL Source Code so the syntax error formatter can show line numbers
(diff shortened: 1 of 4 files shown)


src/main/scala/wdl4s/expression/ValueEvaluator.scala
```diff
@@ -113,7 +113,30 @@ case class ValueEvaluator(override val lookup: String => WdlValue, override val
...
+                              s"""${rhs.getSourceString} is not declared as an output of the sub workflow ${workflowCall.calledWorkflow.fullyQualifiedName}.
+                                 |If you want to use workflow ${workflowCall.calledWorkflow.fullyQualifiedName} as a sub workflow, make sure that its output section is up to date with the latest syntax.
+                                 |See the WDL specification for how to write outputs: https://github.com/broadinstitute/wdl/blob/develop/SPEC.md#outputs""".stripMargin
+                            ))
+                          case taskCall: TaskCall => 
+                            Failure(new WdlExpressionException(
+                              s"""${rhs.getSourceString} is not declared as an output of the task ${taskCall.task.fullyQualifiedName}.
+                                 |Make sure to declare it as an output to be able to use it in the workflow.""".stripMargin
+                            ))
+                          case unkownCall =>
```
mcovarr: unkownCall => unknownCall

src/main/scala/wdl4s/WdlNamespace.scala
```diff
@@ -341,13 +341,22 @@ object WdlNamespace {
...
     )
+    
+    val workflowOutputErrors = workflows flatMap { _.workflowCalls map { _.calledWorkflow } } collect {
+      case workflow if workflow.workflowOutputWildcards.nonEmpty => 
+        new SyntaxError(
+          s"""Workflow ${workflow.unqualifiedName} is used as a sub workflow but has outputs declared with a deprecated syntax not compatible with sub workflows.
+             |To use this workflow as a sub workflow please update the workflow outputs section to the latest WDL specification.
+             |See https://github.com/broadinstitute/wdl/blob/develop/SPEC.md#outputs""".stripMargin
+        )
+    }
```
geoffjentry: I wonder if this could work as a forcomp? I could see why it might not and I didn't try it out, so ...

mcovarr:
:+1:

src/main/scala/wdl4s/expression/ValueEvaluator.scala
```diff
@@ -113,7 +113,30 @@ case class ValueEvaluator(override val lookup: String => WdlValue, override val
               case o: WdlObjectLike =>
                 o.value.get(rhs.getSourceString) match {
                   case Some(v:WdlValue) => Success(v)
-                  case None => Failure(new WdlExpressionException(s"Could not find key ${rhs.getSourceString}"))
+                  case None =>
```
geoffjentry: I don't complain about this as much as I used to but this sub-case is pretty lengthy to follow

geoffjentry:
👍

src/main/scala/wdl4s/expression/ValueEvaluator.scala
Horneth: Yes it is - it's just this whole file is pretty much a giant match case and it seemed even more confusing to me to make a special case method for this and not for the rest, and I didn't want to refactor the entire thing..

PR #74, opened 2017-01-24
Publish hotfix branches using travis.

kshakir:


src/bin/after_success.sh
```diff
@@ -0,0 +1,30 @@
...
+            'set resolvers += Resolver.url("bintray-sbt-plugin-releases", url("http://dl.bintray.com/content/sbt/sbt-plugin-releases"))(Resolver.ivyStylePatterns)' \
+            'set publishTo := Option("artifactory-publish" at "https://artifactory.broadinstitute.org/artifactory/libs-release-local;build.timestamp=" + new java.util.Date().getTime)' \
+            "set credentials += Credentials(\"Artifactory Realm\", \"artifactory.broadinstitute.org\", \"${ARTIFACTORY_USERNAME}\", \"${ARTIFACTORY_PASSWORD}\")" \
+            publish 2>&1 | sed 's/.*set credentials.*/REDACTED LOG LINE/'
+
+    elif [[ "$TRAVIS_BRANCH" =~ ^[0-9\.]+_hotfix$ ]]; then
+        sbt \
+            'set test in Test := {}' \
+            'set isSnapshot := false' \
+            'set git.gitUncommittedChanges := false' \
```
mcovarr: why this?

kshakir: There really should be more cross commenting between the two, but for now...

This is tightly linked to hoops we jump through in our build.sbt. Over in build.sbt, for the default behavior when someone is building, we set this flag to `true`. That triggers the git hashing plugin to append `-SNAP`. Overriding the flag to false here unsets that appending behavior.

kshakir:
Reminder if this PR is merged for use on 0.8-- the 0.8_hotfix branch will need this patch also hotfixed to that branch also, otherwise the code to publish won't actually be there!

geoffjentry:
👍

Horneth:
👍

src/bin/after_success.sh
mcovarr: yeah a comment for posterity might be nice  😄

kshakir:
FYI: This code is not currently needed, as 0.9 was released instead of hotfixing 0.8. Still, merging for now just in case it's needed for a different set of patches.

PR #75, opened 2017-01-24
fix expanded workflow output unqualified name

Horneth:
`output.locallyQualifiedName(this)` resulted in the workflow output unqualified name being `workflow_name.task_name.task_output_name`, which contains the workflow name and hence gives the wrong result when calling `fullyQualifiedName` on it.


ruchim:
:+1:

mcovarr:
:+1:

PR #77, opened 2017-01-26
Harmonize workflow outputs

Horneth:
The last few fixes introduced some incompatibilities between how the workflow outputs look for the old and new syntax. This squares things up and allow Cromwell + other users of wdl4s to get coherent results in both cases.

Workflow outputs unqualified names are now as they should for both old and new syntax: 

- For the old syntax, the unqualified name of the workflow output will be of the form `task_name.output_name`.

```
workflow wf {
  outputs {
    task_name.output_name # will create a WorkflowOutput declaration with "task_name.output_name" as unqualified name
  }
}
```

- For the new syntax, the unqualified name of the workflow output will be the one given in the wdl

```
workflow wf {
  outputs {
    String my_wf_output # will create a WorkflowOutput declaration with "my_wf_output" as unqualified name
  }
}
```

- In order to get a workflow output qualified name containing the workflow name as well, one can 
```
// Assuming workflow is a Workflow object
val outputs = workflow.outputs map { _.locallyQualifiedName(workflow) }
```

- output evaluation function (for task and workflow) now return a `Map[TaskOutput, WdlValue]` and `Map[WorkflowOutput, WdlValue]` respectively. This allows the caller to manipulate the workflow output qualified name as they please for display etc...

- The `WdlArrayType` change is a fix from @cjllanwarne

### DIFF
--- a/src/main/scala/wdl4s/Task.scala
+++ b/src/main/scala/wdl4s/Task.scala
@@ -169,9 +169,9 @@ case class Task(name: String,
 
   def evaluateOutputs(inputs: EvaluatedTaskInputs,
                       wdlFunctions: WdlStandardLibraryFunctions,
-                      postMapper: WdlValue => Try[WdlValue] = v => Success(v)): Try[Map[LocallyQualifiedName, WdlValue]] = {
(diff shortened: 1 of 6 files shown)


cjllanwarne:
👍

ruchim:
:+1:

PR #79, opened 2017-02-02
Files in complex types

cjllanwarne:


kshakir:
👍

src/main/scala/wdl4s/expression/FileEvaluator.scala
```diff
@@ -55,6 +55,8 @@ case class FileEvaluator(valueEvaluator: ValueEvaluator, coerceTo: WdlType = Wdl
         a.value.flatMap(findWdlFiles(_, coerce=false))
       case Success(m: WdlMap) =>
         m.value flatMap { case (k, v) => Seq(k, v) } flatMap(findWdlFiles(_, coerce=false)) toSeq
+      case Success(WdlOptionalValue(_, Some(v))) => findWdlFiles(v)
+      case Success(WdlPair(l, r)) => findWdlFiles(l) ++ findWdlFiles(r)
```
kshakir: ToL: I don't know much about this method. Why do the above set `coerce = false`, but these leave `coerce = true`?

cjllanwarne: Good catch. I believe this should be at the top level only since after coercing a composite type everything inside that needs to be a `File` will already be a `File`. I'll modify this.

src/main/scala/wdl4s/command/ParameterCommandPart.scala
```diff
@@ -25,6 +25,13 @@ object ParameterCommandPart {
     }
     new ParameterCommandPart(attributes, expression)
   }
+
+  private def withPostProcesser(f: WdlValue => WdlValue, valueFunctions: WdlFunctions[WdlValue]) = new WdlFunctions[WdlValue] {
```
ruchim: TOL: It' be nice to have documentation for this method

geoffjentry:
:+1:

ruchim:
:+1:

src/main/scala/wdl4s/command/ParameterCommandPart.scala
katevoss: 👍 for documentation!

PR #81, opened 2017-02-06
Enforce coersion during array construction

cjllanwarne:


ruchim:
:+1:

src/main/scala/wdl4s/values/WdlArray.scala
```diff
@@ -12,25 +13,26 @@ object WdlArray {
       WdlArray(WdlArrayType(WdlStringType), line.split("\t").toSeq.map(WdlString))
     })
   }
-}
 
-case class WdlArray(wdlType: WdlArrayType, value: Seq[WdlValue]) extends WdlValue with TsvSerializable {
+  def apply(wdlType: WdlArrayType, value: Seq[WdlValue]): WdlArray = {
+    if (wdlType == WdlMaybeEmptyArrayType.EmptyArrayType && value.nonEmpty) {
+      throw new UnsupportedOperationException(s"An Array[Void] must be empty but instead has value: ${value.mkString(", ")}")
```
mcovarr: I thought you went with `Nothing` over `Void`

src/main/scala/wdl4s/values/WdlArray.scala
```diff
@@ -12,25 +13,26 @@ object WdlArray {
...
-    throw new UnsupportedOperationException(s"Cannot construct ${wdlType.memberType.toWdlString} array with mixed type: ${value.map(_.wdlType).toSet.mkString(", ")}")
-  }
-  if (wdlType == WdlMaybeEmptyArrayType.EmptyArrayType && value.nonEmpty) {
-    throw new UnsupportedOperationException(s"An Array[Void] must be empty but instead has value: ${value.mkString(", ")}")
-  }
-  if (nonEmpty && value.isEmpty) {
-    throw new UnsupportedOperationException(s"An ${wdlType.toWdlString} contain at least one element")
+    val coercedValue = TryUtil.sequence(value map wdlType.memberType.coerceRawValue)
+    coercedValue match {
+      case Success(coercedArray) => new WdlArray(wdlType, coercedArray) {}
```
mcovarr: is the `new` required?

mcovarr:
:+1:

src/main/scala/wdl4s/values/WdlArray.scala
cjllanwarne: Yes, it makes us use the constructor instead of recursing back into the `apply`

PR #82, opened 2017-02-06
Prioritize declaration expressions above declarations.

mcovarr:
The existing order of lookup resolvers gave the declaration resolver precedence over the declaration expression resolver.  But in the case of an initialized optional variable, the declaration resolver does not fail but rather returns an empty optional.  This non-failing result took precedence over the correct result in the declaration expression resolver.  These changes only flip the precedence of the resolvers.

### DIFF
--- a/src/main/scala/wdl4s/Call.scala
+++ b/src/main/scala/wdl4s/Call.scala
@@ -180,7 +180,7 @@ sealed abstract class Call(val alias: Option[String],
         parentLookup <- Try(parent.lookupFunction(inputs, wdlFunctions, outputResolver, shards, relativeTo)(name))
       } yield parentLookup
 
-      val resolutions = Seq(inputMappingsLookup, declarationLookup, declarationExprLookup, taskParentResolution)
+      val resolutions = Seq(inputMappingsLookup, declarationExprLookup, declarationLookup, taskParentResolution)
 
       resolutions.collectFirst({ case Success(value) => value}) match {
         case Some(value) => value


geoffjentry:
👍 from me but it'd be good to hear from either @cjllanwarne or @Horneth

src/main/scala/wdl4s/Call.scala
```diff
@@ -180,7 +180,7 @@ sealed abstract class Call(val alias: Option[String],
         parentLookup <- Try(parent.lookupFunction(inputs, wdlFunctions, outputResolver, shards, relativeTo)(name))
       } yield parentLookup
 
-      val resolutions = Seq(inputMappingsLookup, declarationLookup, declarationExprLookup, taskParentResolution)
+      val resolutions = Seq(inputMappingsLookup, declarationExprLookup, declarationLookup, taskParentResolution)
```
mcovarr: I'm not sure why this wasn't done with lazy streams to avoid needlessly running lower priority resolvers, but I chose to keep the changes minimal for this PR.

geoffjentry: streams are one of a handful of things that I'm starting to think we'll be seeing a lot more of in wdl4s. another example would be vectors instead of the generic seqs due to all of the `++`s and such.

mcovarr:
Yeah I too would like to hear from our wdl4s experts.  I'm wondering if the problem is actually that the declaration resolver shouldn't succeed for an uninitialized optional.

cjllanwarne:
Incorporated into https://github.com/broadinstitute/wdl4s/pull/81

PR #83, opened 2017-02-10
Fix for Cromwell #1958

ruchim:
Try around the creation of WdlNamespacesWithWorkflow


Horneth:
Should the other `load` methods also return a `Try` for consistency ? Or even `WdlNamespace.loadUsingSource` ?
Otherwise 👍

cjllanwarne:
👍

PR #85, opened 2017-02-13
Defining defined

cjllanwarne:


src/main/scala/wdl4s/expression/WdlStandardLibraryFunctions.scala
```diff
@@ -336,6 +341,8 @@ class WdlStandardLibraryFunctionsType extends WdlFunctions[WdlType] {
     case WdlArrayType(WdlOptionalType(innerType)) => Success(WdlArrayType(innerType))
     case other => Failure(new IllegalArgumentException(s"select_all failed. It expects an array of optional values but got ${other.toWdlString}."))
   }
+  def defined(params: Seq[Try[WdlType]]): Try[WdlType] = extractSingleArgument("defined", params).map(_ => WdlBooleanType)
+
```
kshakir: I have no idea why, but FYI this is the only newline between any of these functions in this class, so I'd say remove it for consistency or add the other missing newlines.

kshakir:
👍

mcovarr:
:+1:

PR #87, opened 2017-02-22
Fixes for glob indexing.

kshakir:
In the FileEvaluator, for arrays/maps, evaluate files from the left and right side.
For binary operators, don't evaluate the right if the left fails.

### DIFF
--- a/src/main/scala/wdl4s/expression/FileEvaluator.scala
+++ b/src/main/scala/wdl4s/expression/FileEvaluator.scala
@@ -98,10 +98,10 @@ case class FileEvaluator(valueEvaluator: ValueEvaluator, coerceTo: WdlType = Wdl
       case a: Ast if a.isBinaryOperator =>
         evalValueToWdlFile(a) match {
           case Success(f: WdlFile) => Success(Seq(f))
-          case _ => (evaluateRecursive(a.getAttribute("lhs")), evaluateRecursive(a.getAttribute("rhs"))) match {
-            case (Success(a:Seq[WdlFile]), Success(b:Seq[WdlFile])) => Success(a ++ b)
-            case _ => Failure(new WdlExpressionException(s"Could not evaluate:\n${a.toPrettyString}"))
-          }
+          case _ => for {
+            left <- evaluateRecursive(a.getAttribute("lhs"))
+            right <- evaluateRecursive(a.getAttribute("rhs"))
+          } yield left ++ right
         }
       case a: Ast if a.isUnaryOperator =>
         evalValueToWdlFile(a) match {
@@ -115,7 +115,10 @@ case class FileEvaluator(valueEvaluator: ValueEvaluator, coerceTo: WdlType = Wdl
       case a: Ast if a.isArrayOrMapLookup =>
         evalValue(a) match {
           case Success(f: WdlFile) => Success(Seq(f))
-          case _ => evaluateRecursive(a.getAttribute("rhs"))
+          case _ => for {
+            left <- evaluateRecursive(a.getAttribute("lhs"))
+            right <- evaluateRecursive(a.getAttribute("rhs"))
+          } yield left ++ right
         }
       case a: Ast if a.isMemberAccess =>
         evalValue(a) match {
--- a/src/test/scala/wdl4s/expression/FileEvaluatorSpec.scala
+++ b/src/test/scala/wdl4s/expression/FileEvaluatorSpec.scala
@@ -1,6 +1,6 @@
 package wdl4s.expression
 
-import wdl4s.WdlExpression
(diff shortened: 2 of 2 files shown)


src/main/scala/wdl4s/expression/FileEvaluator.scala
```diff
@@ -115,7 +115,7 @@ case class FileEvaluator(valueEvaluator: ValueEvaluator, coerceTo: WdlType = Wdl
       case a: Ast if a.isArrayOrMapLookup =>
         evalValue(a) match {
           case Success(f: WdlFile) => Success(Seq(f))
-          case _ => evaluateRecursive(a.getAttribute("rhs"))
+          case _ => evaluateRecursive(a.getAttribute("lhs"))
```
mcovarr: If there's an expression on the rhs that's no longer going to be evaluated?  Why don't we want to evaluate both sides?

mcovarr: e.g. `File globFile = glob("globFile.txt")[count - 1]`

kshakir: Based on the comment [above](https://github.com/broadinstitute/wdl4s/blob/e040cf832588aea71fd5e4eaa0e05a4b397c4bd2/src/main/scala/wdl4s/expression/FileEvaluator.scala#L72-L87), I believe this method only supposed to be used during early evaluation, _before_ a task actually runs.

So `glob("globFile.txt")[count - 1]` should return a `Seq(WdlGlobFile("globFile.txt"))` for this method / case. Right?

mcovarr: shrug?  I'd like to know Centaur's opinion on this. 🙂

kshakir: Centaur last night said it was fine. You'll have to take my offline word for it though, as- as far as I know- we don't have a way of travis testing wdl4s PRs.

geoffjentry: Not at the moment at least. I've been going with the hope, pray, be ready to revert if necessary method.

geoffjentry: Also @kshakir I'm buying what you're selling. But I reserve the right to say that I sided with @mcovarr all along if travis winds up irritated!

geoffjentry:
@kshakir is there a ticket associated w/ this? I'm mainly just interested to know the background

geoffjentry:
👍

src/main/scala/wdl4s/expression/FileEvaluator.scala
Horneth: The only think I can think of that not evaluating the `rhs` would do is fail to detect if someone uses files as keys for a map...
If we really want to catch this case we could still do something like
```scala
evaluateRecursive(a.getAttribute("lhs")) flatMap { lhs => evaluateRecursive(a.getAttribute("rhs")) map { rhs => lhs ++ rhs } }
```

Horneth:
👍

kshakir:
Updated to evaluate left and right sides of map/array lookups.

src/test/scala/wdl4s/expression/FileEvaluatorSpec.scala
```diff
@@ -45,12 +42,13 @@ class FileEvaluatorSpec extends FlatSpec with Matchers {
     (""" [read_string("x"), read_string("y")] """, Seq(WdlSingleFile("x"), WdlSingleFile("y")), WdlArrayType(WdlStringType)),
     (""" {read_int("a"): read_string("x"), 4: read_string("y")} """, Seq(WdlSingleFile("a"), WdlSingleFile("x"), WdlSingleFile("y")), WdlArrayType(WdlStringType)),
     (""" glob("out-*.txt") """, Seq(WdlGlobFile("out-*.txt")), WdlFileType),
+    (""" glob("out-*.txt")[0] """, Seq(WdlGlobFile("out-*.txt")), WdlFileType),
```
mcovarr: can you have an `rhs` array or map indexing test here?

mcovarr:
:+1: anew

src/test/scala/wdl4s/expression/FileEvaluatorSpec.scala
kshakir: I don't know of any that would actually pass parsing or generate a proper AST. Filing a new ticket.

kshakir: https://github.com/broadinstitute/wdl4s/issues/90

PR #91, opened 2017-02-26
Various updates.

kshakir:
Switched `WdlInteger` value from java reference to scala value.
Slightly better error messages when trying to coerce `Array[Int]?` to `Array[Int?]`, or vice versa.
Cleaned up scaladoc warnings.

### DIFF
--- a/src/main/scala/wdl4s/Scope.scala
+++ b/src/main/scala/wdl4s/Scope.scala
@@ -182,7 +182,7 @@ trait Scope {
     * @param wdlFunctions Implementation of WDL functions for expression evaluation
     * @param shards For resolving specific shards of scatter blocks
     * @return String => WdlValue lookup function rooted at `scope`
-    * @throws VariableNotFoundException => If no errors occurred, but also `name` didn't resolve to any value
+    * @throws VariableNotFoundException If no errors occurred, but also `name` didn't resolve to any value
     * @throws VariableLookupException if anything else goes wrong in looking up a value for `name`
     */
   def lookupFunction(knownInputs: WorkflowCoercedInputs,
@@ -204,7 +204,7 @@ trait Scope {
           }
         case (Success(value: WdlArray), Some(shard)) =>
           Failure(new VariableLookupException(s"Scatter expression (${scatter.collection.toWdlString}) evaluated to an array of ${value.value.size} elements, but element $shard was requested."))
-        case (Success(value: WdlArray), None) =>
+        case (Success(_: WdlArray), None) =>
           Failure(ScatterIndexNotFound(s"Could not find the shard mapping to this scatter ${scatter.fullyQualifiedName}"))
         case (Success(value: WdlValue), _) =>
           Failure(new VariableLookupException(s"Expected scatter expression (${scatter.collection.toWdlString}) to evaluate to an Array.  Instead, got a $value"))
--- a/src/main/scala/wdl4s/types/WdlType.scala
+++ b/src/main/scala/wdl4s/types/WdlType.scala
@@ -6,6 +6,7 @@ import wdl4s.{WdlExpressionException, WdlSource, WdlSyntaxErrorFormatter}
 import wdl4s.parser.WdlParser
 
 import scala.collection.JavaConverters._
+import scala.runtime.ScalaRunTime
 import scala.util.{Failure, Success, Try}
 
 class WdlTypeException(message: String) extends RuntimeException(message)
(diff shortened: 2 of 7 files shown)


kshakir:
FYI for reviewers-- In addition to tests in this repo, tests pass in both cromwell `sbt alltests:test` and centaur `sbt test`.

src/main/scala/wdl4s/values/WdlInteger.scala
```diff
@@ -5,7 +5,7 @@ import wdl4s.types.WdlIntegerType
 
 import scala.util.{Failure, Success, Try}
 
-case class WdlInteger(value: Integer) extends WdlPrimitive {
+case class WdlInteger(value: Int) extends WdlPrimitive {
```
geoffjentry: Just curious what led to this, or was it just something you noticed?

kshakir: This is a spinoff of https://github.com/broadinstitute/cromwell/pull/2031

In this IntelliJ screenshot:
toInt implicit underlined

That implicit underlining the `.toInt` on line 178 shouldn't be there.

src/main/scala/wdl4s/Scope.scala
```diff
@@ -181,8 +181,8 @@ trait Scope {
     * @param knownInputs All known values of FQNs
     * @param wdlFunctions Implementation of WDL functions for expression evaluation
     * @param shards For resolving specific shards of scatter blocks
-    * @return String => WdlValue lookup function rooted at `scope`
```
mcovarr: why was this rocket removed?

src/main/scala/wdl4s/types/WdlType.scala
```diff
@@ -28,8 +29,12 @@ trait WdlType {
    */
   def coerceRawValue(any: Any): Try[WdlValue] = {
     any match {
-      case v: WdlValue if v.wdlType == this => Success(v)
-      case a if !coercion.isDefinedAt(any) => Failure(new IllegalArgumentException(s"No coercion defined from '$any' of type '${any.getClass}' to ${getClass.getSimpleName}."))
+      case wdlValue: WdlValue if wdlValue.wdlType == this => Success(wdlValue)
+      case wdlValue: WdlValue if !coercion.isDefinedAt(any) => Failure(new IllegalArgumentException(
+        s"No coercion defined from '${ScalaRunTime.stringOf(any, 3)}' of type '${wdlValue.wdlType}' to '$this'."))
```
mcovarr: what is this `ScalaRunTime.stringOf` doing?  I'm not able to find docs for this.  😦

mcovarr:
:+1:

src/main/scala/wdl4s/Scope.scala
kshakir: Good catch. Overly aggressive removal. The one on the next line was causing bad scaladoc rendering [due to](https://wiki.scala-lang.org/display/SW/Syntax):
> Headings: =heading= defines a heading. Lower-level heading are obtained by using more = signs, like ===sub-heading===. A heading must be define on its own line.

producing:
```bash
[warn] /Users/kshakir/src/wdl4s/src/main/scala/wdl4s/Scope.scala:176: unbalanced or unclosed heading
[warn]   /**
[warn]   ^
```

Apparently `@return` doesn't render the same way as `@throws`, and the former rocket shows up just fine in the rendered html.

Will put it back in a squash.

src/main/scala/wdl4s/types/WdlType.scala
kshakir: Hehe, good catch also. `scala.runtime.ScalaRuntime` is a lot like `com.sun`, aka "this is undocumented, use this and don't complain if you get burned": https://github.com/scala/scala/blob/v2.11.8/src/library/scala/runtime/ScalaRunTime.scala#L23-L26

But it does exactly what I want, which is to properly render most `Any`s, including arrays-to-string. I didn't like the other hand-crafted suggestions in [this stackoverflow thread](http://stackoverflow.com/questions/3328085/scala-printing-arrays/3328714#3328714).

kshakir:
Re-review please. See latest commit with changes in the error messages. Hopefully they're much simpler, but still contain enough detail for a human to debug wth is wrong.

mcovarr:
:+1: nice, both the code and the error messages are much easier to read! 😄

src/main/scala/wdl4s/types/WdlType.scala
```diff
@@ -28,8 +29,13 @@ trait WdlType {
    */
   def coerceRawValue(any: Any): Try[WdlValue] = {
     any match {
-      case v: WdlValue if v.wdlType == this => Success(v)
-      case a if !coercion.isDefinedAt(any) => Failure(new IllegalArgumentException(s"No coercion defined from '$any' of type '${any.getClass}' to ${getClass.getSimpleName}."))
+      case wdlValue: WdlValue if wdlValue.wdlType == this => Success(wdlValue)
+      case wdlValue: WdlValue if !coercion.isDefinedAt(any) => Failure(new IllegalArgumentException(
+        s"No coercion defined from '${wdlValue.toWdlString}' of type" +
```
cjllanwarne: I personally don't like including the value here. It's almost always unreadably long and not very useful in debugging...

cjllanwarne:
👍 otherwise

src/main/scala/wdl4s/types/WdlType.scala
cjllanwarne: (ToL)

kshakir:
One more re-review please. For the coercion errors printed out, the latest commit limits the number of elements per collection to 3.

src/main/scala/wdl4s/values/WdlValue.scala
```diff
@@ -38,7 +37,7 @@ trait WdlValue {
   def typeName: String = wdlType.getClass.getSimpleName
 
   /* This emits valid WDL source.  WdlString("foobar") -> "foobar" (quotes included) */
-  def toWdlString: String = ???
+  def toWdlString: String = throw new NotImplementedError(s"$getClass does not implement toWdlString")
```
cjllanwarne: ToL (no change req'd) - that comment on :39 seems somewhat optimistic!

cjllanwarne:
Consider that request for change a consideration/ToL, but I do think at least something to indicate truncation (even a "ps, that might have been truncated" in the message) would be really useful to avoid confusion

kshakir:
The ellipses identifying truncation are something I've thought about, and would be a good addition, but I don't have a quick fix for. Going to punt that for the future, as this error is appearing for an already instantiated collection, where a coercion has failed based on the expected type. The identification of the contents of the wdlValue are more hints to the user of "it was _this_ Array[Int?] that failed coercion, versus _that_ Array[Int?]".

cjllanwarne:
re-re-👍

mcovarr:
:+1: :+1: :+1:

PR #96, opened 2017-03-25
2.12 support plus cross-version artifact pushes

geoffjentry:


build.sbt
```diff
@@ -25,30 +27,29 @@ versionWithGit ++ versionSettings
...
   "io.spray" %% "spray-json" % sprayJsonV,
   /*
   Exclude test framework cats-laws and its transitive dependency scalacheck.
   If sbt detects scalacheck, it tries to run it.
   Explicitly excluding the two problematic artifacts instead of including the three (or four?).
   https://github.com/typelevel/cats/tree/v0.7.2#getting-started
    */
-  "org.typelevel" %% "cats" % "0.7.2"
+  "org.typelevel" %% "cats" % "0.9.0"
     exclude("org.typelevel", "cats-laws_2.11")
```
geoffjentry: @kshakir I just noticed this (would exist on centaur as well) - should I throw in a 2.12 line for these as well?

Horneth:
👍

build.sbt
```diff
@@ -25,33 +27,37 @@ versionWithGit ++ versionSettings
...
-  "commons-io" % "commons-io" % "2.5",
-  "org.apache.commons" % "commons-lang3" % "3.4",
-  "com.github.pathikrit" %% "better-files" % "2.16.0",
-  //---------- Test libraries -------------------//
-  "org.scalatest" %% "scalatest" % "3.0.1" % Test,
-  "org.pegdown" % "pegdown" % "1.6.0" % Test
-)
+libraryDependencies ++= {
+  // the if doesn't work if we put this up with the other fooV's
+  val betterFilesV = if (scalaBinaryVersion.value == "2.11") "2.17.1" else "3.0.0"
```
kshakir: ToL: Ew. That looks brittle.

IMHO, I'd love for better.files to just _not_ be used in wdl4s. It leaks this downstream dependency complexity into any other component that wants to use wdl4s. Meanwhile, `java.nio` is universal across scala versions, and actually good enough for the three small places better.files is used in the wdl4s main code.

Similarly, I don't like that better.files is [exposed](https://github.com/broadinstitute/wdl4s/blob/8d0ac9f66d44aa0fea67ccc299d91ce0618e3461/src/main/scala/wdl4s/WdlNamespace.scala#L542) in wdl4s public apis.

kshakir:
better.files _feels_ like it's going to cause someone upgrade pain later, but for now, LGTM 👍

build.sbt
geoffjentry: We can make it 2.17.1 for both here, I just thought it was neat that you can split versions. It might also be the case at some point that we'll need to take advantage of this trick at some point. But at the moment it's not necessary if you'd rather not see it.

kshakir: Interesting. I'm fine leaving this as is, and only ever modifying if/when necessary.

geoffjentry: I already got rid of it ;)

PR #97, opened 2017-03-26
if then else expressions!

cjllanwarne:
This is hopefully a really handy piece of syntactic sugar to avoid erstwhile WDL-value-gymnastics:
```
Boolean b = ...

Int memory = if b then "2GB" else "5GB"
String docker = if b then "ubuntu:latest" else "my:docker"
```


geoffjentry:
Make sure the syntax has been vetted by the usual suspects - @katevoss @vdauwera @yfarjoun etc. Don't rely on them to have been reading their github notifications either :)

cjllanwarne:
@geoffjentry absolutely 100x yes.  This actually started as a weekend "I wonder how difficult..." that subsequently turned into a "hey actually this could be really easy..."

src/test/scala/wdl4s/expression/ValueEvaluatorSpec.scala
```diff
@@ -512,4 +512,16 @@ class ValueEvaluatorSpec extends FlatSpec with Matchers {
     val expectedMessage = "Sorry! Operation + is not supported on empty optional values. You might resolve this using select_first([optional, default]) to guarantee that you have a filled value."
     hello.add(noneWorld) should be(Failure(OptionalNotSuppliedException("+")))
   }
+
+  "Ternary if blocks" should "evaluate LHS if the condition is true" in {
+    constEval(""" if (5 == 4 + 1) then 6 + 7 else 14 * 15 """) should be(WdlInteger(13))
```
yfarjoun: since I was @ summoned...

would it be appropriate to check that only one of these expressions is actually **evaluated** (rather than only checking that the return value is correct)

yfarjoun:
can you also do things like 

if condition then call task1 else call task2 ?

meganshand:
@cjllanwarne This looks so helpful! Thank you!

Could you also do the following:

```
Boolean b = True
Int small_memory = 2
Int large_memory = 5

Int memory = if b then small_memory else large_memory
```

Where presumably you could declare `b`, `small_memory`, and `large_memory` either in the WDL itself or in the json?

Also, I'm assuming based on your comment that this would only apply to declaring variables and not to calling tasks, which of course in the future would also be helpful.

Horneth:
A test or 2 for type evaluation and file evaluation that would be nice.
Otherwise this is very cool ! 👍

mcovarr:
:+1: with a test like what Yossi suggested

src/test/scala/wdl4s/expression/ValueEvaluatorSpec.scala
cjllanwarne: done.

src/test/scala/wdl4s/expression/FileEvaluatorSpec.scala
```diff
@@ -40,15 +40,26 @@ class FileEvaluatorSpec extends FlatSpec with Matchers {
...
+    (""" [fileNameAsStringInput, "${fileNameAsStringInput}.bai"] """, Seq(WdlSingleFile("sommat.bam"), WdlSingleFile("sommat.bam.bai")), WdlArrayType(WdlFileType)),
+    (""" [ fileNameAsStringInput, mapToFileName["Chris"] ] """, Seq(WdlSingleFile("sommat.bam"), WdlSingleFile("sommatStupid.bam")), WdlArrayType(WdlFileType)),
+    (""" {read_int("a"): read_string("x"), 4: read_string("y")} """, Seq(WdlSingleFile("a"), WdlSingleFile("x"), WdlSingleFile("y")), WdlMapType(WdlIntegerType, WdlStringType)),
     (""" glob("out-*.txt") """, Seq(WdlGlobFile("out-*.txt")), WdlFileType),
     (""" glob("out-*.txt")[0] """, Seq(WdlGlobFile("out-*.txt")), WdlFileType),
-    (""" read_tsv("my_file") """, Seq(WdlSingleFile("my_file")), WdlFileType)
+    (""" read_tsv("my_file") """, Seq(WdlSingleFile("my_file")), WdlFileType),
+    (""" if read_int("i.txt") == 10 then "a.txt" else "b.txt" """, Seq(WdlSingleFile("a.txt"), WdlSingleFile("b.txt"), WdlSingleFile("i.txt")), WdlFileType),
+    (""" if "a" == "b" then "a.txt" else "b.txt" """, Seq(WdlSingleFile("b.txt")), WdlFileType),
```
Horneth: How does it know that "a.txt" and "b.txt" are files here ?

cjllanwarne: It's the final `WdlFileType` - which in a real output declaration would be passed in based on the declared type of the output

PR #108, opened 2017-05-13
Allow Map to Array coercion

cjllanwarne:
Allows, e.g.:
```
task print {
	Int id
	String value
	command {
		echo ${id}: ${value}
	}
	output {
		String out = read_string(stdout())
	}
}


workflow foo {
	Map[Int, String] map  = {
		1: "one",
		2: "two",
		3: "three"
	}

	Array[Pair[Int, String]] listOfPairs = map
	scatter (pair in listOfPairs) {
	  call print {input: id = pair.left, value = pair.right}
	}
}
```

### DIFF
--- a/src/main/scala/wdl4s/types/WdlArrayType.scala
+++ b/src/main/scala/wdl4s/types/WdlArrayType.scala
@@ -1,8 +1,9 @@
 package wdl4s.types
 
 import lenthall.util.TryUtil
-import wdl4s.values.{WdlArray, WdlFile, WdlString, WdlValue}
+import wdl4s.values.{WdlArray, WdlFile, WdlMap, WdlPair, WdlString, WdlValue}
 import spray.json.JsArray
+import wdl4s.values.WdlArray.WdlArrayLike
 
 import scala.collection.JavaConverters._
 import scala.util.{Failure, Success}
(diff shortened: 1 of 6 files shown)


geoffjentry:
Oh interesting!

I ask this more out of my own curiosity than any amount of skepticism but was this something you've seen asked for or did it just seem like a good idea?

Also, while I can't see why it'd be controversial, it's worth passing it past ye olde focus group

src/main/scala/wdl4s/types/WdlArrayType.scala
```diff
@@ -15,26 +15,33 @@ sealed trait WdlArrayType extends WdlType {
...
     case s:Seq[Any] if s.nonEmpty =>
       val coerced = s.map {memberType.coerceRawValue(_).get}
-      WdlArray(WdlArrayType(coerced.head.wdlType), coerced)
-    case _ => WdlArray(WdlArrayType(memberType), Seq())
+      WdlArray(this, coerced)
+    case _ => WdlArray(this, Seq())
   }
 
-  protected def coercionMaker(allowEmpty: Boolean): PartialFunction[Any, WdlValue] = {
+  private val allowEmpty = !guaranteedNonEmpty
```
geoffjentry: Not really keen on this whole `guaranteedNonEmpty` thing but it seems to be a preexisting condition and those aren't covered anymore so consider this a TOL

cjllanwarne: You don't like the name, or the concept?

geoffjentry: The latter. IOW why is it not baked into the type system.

cjllanwarne: If I understand you correctly, "this should be a method in the sealed trait, not a property of the subclasses"

geoffjentry: that'd be one way to do it, if i understood *you* correctly. but it's not a big deal

Horneth:
I think if we replace this https://github.com/broadinstitute/cromwell/blob/develop/engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala#L541 (and maybe in a few other places) with a check that it can be coerced to an array instead we will allow directly 

```
workflow foo {
	Map[Int, String] map  = {
		1: "one",
		2: "two",
		3: "three"
	}

	scatter (pair in map) {
	  call print {input: id = pair.left, value = pair.right}
	}
}
```

Horneth:
👍

cjllanwarne:
@Horneth nice! I'll add that when I bring this into Cromwel

ruchim:
:+1: LGTM

PR #110, opened 2017-05-14
Validate string passed to WdlSingleFile

jsotobroad:
I wasn't sure where I should put other tests for this or include more in wdl4s itself.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -36,7 +36,9 @@ resolvers ++= List(
 libraryDependencies ++= {
   Seq(
     "org.broadinstitute" %% "lenthall" % lenthallV,
+    "com.google.guava" % "guava" % "20.0",
     "com.typesafe.scala-logging" %% "scala-logging" % "3.5.0",
+    "eu.timepit" %% "refined" % "0.7.0",
     "io.spray" %% "spray-json" % sprayJsonV,
     /*
     Exclude test framework cats-laws and its transitive dependency scalacheck.
--- a/src/main/scala/wdl4s/values/WdlFile.scala
+++ b/src/main/scala/wdl4s/values/WdlFile.scala
@@ -1,11 +1,15 @@
 package wdl4s.values
 
+import com.google.common.net.UrlEscapers
 import wdl4s.types.{WdlFileType, WdlType}
+import eu.timepit.refined.refineV
+import eu.timepit.refined.string.Uri
+import wdl4s.exception.UnsatisfiedInputException
 
 import scala.util.{Success, Try}
 
 object WdlFile {
-  def appendPathsWithSlashSeparators(path1: String, path2: String) = {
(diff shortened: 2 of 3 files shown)


build.sbt
```diff
@@ -36,7 +36,9 @@ resolvers ++= List(
 libraryDependencies ++= {
   Seq(
     "org.broadinstitute" %% "lenthall" % lenthallV,
+    "com.google.guava" % "guava" % "20.0",
```
geoffjentry: Why are you pulling in guava?

src/main/scala/wdl4s/values/WdlFile.scala
```diff
@@ -35,14 +39,20 @@ sealed trait WdlFile extends WdlPrimitive {
...
-  override def valueString = value.toString
+  override def valueString: String = value.toString
 }
 
-case class WdlSingleFile(value: String) extends WdlFile {
-  override def toWdlString = "\"" + value.toString + "\""
+case class WdlSingleFile(origValue: String) extends WdlFile {
+  override def toWdlString: String = "\"" + value.toString + "\""
+
+  // Validate file path
```
geoffjentry: You're not really taking advantage of Refined here, at least I don't think so. 

The power of that library is to define a type which has been refined further, e.g. "an integer larger than 5", or "a string which is not empty". You're not actually doing that here.

build.sbt
jsotobroad: I was trying to pull in the library that Cromwell uses to escape strings when creating`URI`s from them just so they had the same behavior.

src/main/scala/wdl4s/values/WdlFile.scala
jsotobroad: I thought this was refining something like "a string that is a valid URI"

cjllanwarne: I think in this case the `origValue` could become refined, e.g. `case class WdlSingleFile(origValue: String Refined UrI)`

geoffjentry: How is that different than an URI?

cjllanwarne: If I understood @jsotobroad correctly, the URI type contains a whole bunch of baggage and difficulty in interpretation. 

I liked the idea of adding the `Refined` type since it lets us check that the string is valid, but when we want to use it, it's still just a String object and is just as easy to manipulate.

Horneth: @geoffjentry Yeah we realized that making it an actual URI, although good on paper turns out to be hard because of the constraints that it puts on the String that we don't necessarily want to enforce (spaces need to be encoded, no underscore in host names etc...)

PR #113, opened 2017-05-30
Input Pair objects through inputs json

ruchim:


src/test/scala/wdl4s/types/WdlPairTypeSpec.scala
```diff
@@ -96,7 +115,34 @@ class WdlPairTypeSpec extends FlatSpec with Matchers {
         |  }
         |}
       """.stripMargin
-    
+
     noException should be thrownBy WdlNamespaceWithWorkflow.load(wdl, Seq.empty)
   }
+
+  it should "coerce a Map into a WdlArray of WdlPair type" in {
```
cjllanwarne: Is this just ypu tidying up after my "map to array of pairs" PR? Thanks! 🙇

src/main/scala/wdl4s/types/WdlPairType.scala
```diff
@@ -1,11 +1,15 @@
...
+import spray.json.JsObject
 import wdl4s.values.{WdlPair, WdlValue}
 
+import scala.language.postfixOps
+
 case class WdlPairType(leftType: WdlType, rightType: WdlType) extends WdlType {
 
   override def isCoerceableFrom(otherType: WdlType): Boolean = otherType match {
     case WdlPairType(otherType1, otherType2) => leftType.isCoerceableFrom(otherType1) && rightType.isCoerceableFrom(otherType2)
+    case jsObject: JsObject if jsObject.fields.size.equals(1) => rightType.coerceRawValue(jsObject.fields.head._2).isSuccess
```
cjllanwarne: This logic seems kinda weird?
Also, `otherType` is a `WdlType` so will never be `JsObject`. Presumably you don't need a case for this here.

src/test/scala/wdl4s/types/WdlPairTypeSpec.scala
```diff
@@ -96,7 +115,34 @@ class WdlPairTypeSpec extends FlatSpec with Matchers {
...
+      case Success(array) => array shouldEqual arrayOfPairs
+      case Failure(f) => fail(s"exception while coercing JsObject: $f")
+    }
+  }
+
+  it should "coerce a complex JsArray into a WdlArray of WdlPairs of WdlArrays" in {
+    val complexJsArray =  JsArray(JsObject("a" -> JsArray(JsNumber(1), JsNumber(11))),
+                                  JsObject("b" -> JsArray(JsNumber(2), JsNumber(21))),
+                                  JsObject("c" -> JsArray(JsNumber(3), JsNumber(31))))
+
```
cjllanwarne: I think we wanted this to be an object with `left` and `right` fields, so that we can have also complex types on the LHS too, e.g. 
```
pairs: [
  {
    left: "a",
    right: [1, 11]
  },
  {
    left: "b",
    right: [2, 22]
  },
  {
    left: "c",
    right: [3, 33]
  },
]
```

src/main/scala/wdl4s/types/WdlPairType.scala
```diff
@@ -19,6 +24,26 @@ case class WdlPairType(leftType: WdlType, rightType: WdlType) extends WdlType {
   override protected def coercion: PartialFunction[Any, WdlValue] = {
     case otherPair @ WdlPair(otherValue1, otherValue2) if isCoerceableFrom(otherPair.wdlType) =>
       WdlPair(leftType.coerceRawValue(otherValue1).get, rightType.coerceRawValue(otherValue2).get)
+    case jsObject: JsObject if jsObject.fields.size == 2 => mapToPair(jsObject.fields, this)
+  }
+
+  def mapToPair(m: Map[String, JsValue], wdlPairType: WdlPairType): WdlPair = {
+
+    val wdlPair: Seq[Try[WdlValue]] = (m.get("Left"), m.get("Right")) match {
```
Horneth: Can we make this case insensitive and allow "left" and "right" too ?

src/main/scala/wdl4s/types/WdlPairType.scala
```diff
@@ -19,6 +24,26 @@ case class WdlPairType(leftType: WdlType, rightType: WdlType) extends WdlType {
...
+      case (Some(leftVal), Some(rightVal)) => Seq(wdlPairType.leftType.coerceRawValue(leftVal), wdlPairType.rightType.coerceRawValue(rightVal))
```
Horneth: Shouldn't there be another `case _ => ...` to account for when it's not a valid json ? (and associated test case)

src/test/scala/wdl4s/types/WdlPairTypeSpec.scala
```diff
@@ -96,7 +122,49 @@ class WdlPairTypeSpec extends FlatSpec with Matchers {
...
+      case Failure(f) => fail(s"exception while coercing JsObject: $f")
+    }
+  }
+
+  it should "coerce a complex JsArray into a WdlArray of WdlPairs of WdlArrays" in {
+    val complexJsArray =  JsArray(
+      JsObject("Left" -> JsArray(JsString("a"), JsString("b")), "Right" -> JsArray(JsNumber(1), JsNumber(11))),
+      JsObject("Left" -> JsArray(JsString("c"), JsString("d")), "Right" -> JsArray(JsNumber(2), JsNumber(21))),
+      JsObject("Left" -> JsArray(JsString("e"), JsString("f")), "Right" -> JsArray(JsNumber(3), JsNumber(31)))
+    )
```
Horneth: Now that it's done don't bother, but it might be easier to write/read to do actual json and just parse it.

src/main/scala/wdl4s/types/WdlPairType.scala
ruchim: When you say valid json-- do you mean it doesn't have the "left" or "right" fields? If yes, then there should be a test case for it already.

src/main/scala/wdl4s/types/WdlPairType.scala
ruchim: Done

src/main/scala/wdl4s/types/WdlPairType.scala
Horneth: yep that's what I meant

Horneth: I think as it is now you'll get a scala match error if you give a json like this

```
{
  "NotRight": "right",
  "NotLeft:" "left"
}
```

src/test/scala/wdl4s/types/WdlPairTypeSpec.scala
cjllanwarne: ToL: Real JSON also has the added benefit of being almost pseudo-documentation in a TDD kind of a way.

src/main/scala/wdl4s/types/WdlPairType.scala
```diff
@@ -19,6 +24,28 @@ case class WdlPairType(leftType: WdlType, rightType: WdlType) extends WdlType {
...
+    }
+
+    val failures = wdlPair collect { case f:Failure[_] => f }
+
+    failures match {
+      case f: Iterable[Failure[_]] if f.nonEmpty =>
+        throw new UnsupportedOperationException(s"Failed to coerce one or more values for creating a ${wdlPairType.toWdlString}:\n${TryUtil.stringifyFailures(f)}}")
+      case _ =>
+        val coercedPair = wdlPair match { case Seq(left, right) => (left.get, right.get) }
```
cjllanwarne: ToL: This `match` is also a bit convoluted. I think what you really want is just a one-liner, `WdlPair(wdlPair._1.get, wdlPair._2.get)`

EDIT: but this requires `wdlPair` to be a tuple. Which then ruins the `wdlPair.collect`. Sigh. If you want, there's a library that can flip tuples into a List, but we might be getting much further away from simple review changes at this point. From [this post](https://stackoverflow.com/questions/1919009/are-there-any-methods-included-in-scala-to-convert-tuples-to-lists):

> Using Shapeless -
```
@ import syntax.std.tuple._
import syntax.std.tuple._
@ (1,2,3).toList
res21: List[Int] = List(1, 2, 3)
@ (1,2,3,4,3,3,3,3,3,3,3).toList
res22: List[Int] = List(1, 2, 3, 4, 3, 3, 3, 3, 3, 3, 3)
```

src/test/scala/wdl4s/types/WdlPairTypeSpec.scala
```diff
@@ -96,7 +122,58 @@ class WdlPairTypeSpec extends FlatSpec with Matchers {
...
+      JsObject("Left" -> JsString("c"), "Right" -> JsNumber(3))
+    )
+
+    WdlArrayType(WdlPairType(WdlStringType, WdlIntegerType)).coerceRawValue(jsArray) match {
+      case Success(array) => array shouldEqual arrayOfPairs
+      case Failure(f) => fail(s"exception while coercing JsObject: $f")
+    }
+  }
+
+  it should "coerce a complex JsArray into a WdlArray of WdlPairs of WdlArrays" in {
```
cjllanwarne: ToL: Making it the outermost element an array just:
- Makes the test harder to read
- Tests that WdlArray is correctly delegating to us

So, I'd be just as confident in this test if it just coerced a single element.

cjllanwarne:
I'd like one more test:
- `it should "report both errors if both sides of an input pair are the wrong type"`
 - I.e. Trying to use a `(String, Int)` input for a `Pair[Int, String]`

Also, if we could document this somehow that'd be great - we've already had comments about lack of docs about referencing the left/right side of a Pair, so inputting in this format is probably worth spelling out somewhere. OTOH if input format is completely undocumented... well maybe just add it as a new ticket.

Otherwise... 👍

src/test/scala/wdl4s/types/WdlPairTypeSpec.scala
ruchim: Changed to json for reals

src/main/scala/wdl4s/types/WdlPairType.scala
```diff
@@ -19,6 +24,28 @@ case class WdlPairType(leftType: WdlType, rightType: WdlType) extends WdlType {
...
+        val leftType = WdlType.homogeneousTypeFromValues(Seq(coercedPair._1))
+        val rightType = WdlType.homogeneousTypeFromValues(Seq(coercedPair._2))
```
ruchim: crud from debugging--thanks for catching that!

ruchim:
@cjllanwarne a PR for pair docs: https://github.com/broadinstitute/wdl/pull/114

Horneth:
👍

PR #120, opened 2017-06-21
bring in cwl parser

danbills:
Parses 1st-tool.cwl and 1st-workflow.cwl.

This is still pretty messy in terms of layout (see model.scala) but it gets the parsing job done.

For reference:
http://www.commonwl.org/v1.0/Workflow.html


mcovarr:
This fails to compile for me for the same reasons it's failing to compile in Travis.

mcovarr:
nm, that last commit fixed the compilation error.

build.sbt
```diff
@@ -51,11 +54,31 @@ libraryDependencies ++= {
...
+    "io.circe" %% "circe-yaml" % "0.6.1",
+
+    "eu.timepit" %% "refined"            % "0.8.2",
+    "com.github.benhutchison" %% "mouse" % "0.8",
+
+    "org.scalatest" %% "scalatest" % "3.0.1" % "test",
+    "com.lihaoyi" %% "ammonite-ops" % "1.0.0-RC7" % "test",
+    "org.scalatest" %% "scalatest" % "3.0.1" % "test",
+
+    "com.beachape" %% "enumeratum" % enumeratumVersion,
```
danbills: It turns out the parsing works with plain ol' Scala enumerations, so I plan on removing this

build.sbt
```diff
@@ -51,11 +54,31 @@ libraryDependencies ++= {
...
     "org.scalatest" %% "scalatest" % "3.0.1" % Test,
     "org.pegdown" % "pegdown" % "1.6.0" % Test
   )
 }
 
+libraryDependencies ++= Seq(
+  "io.circe" %% "circe-generic",
+  "io.circe" %% "circe-shapes",
+  "io.circe" %% "circe-refined"
+).map(_ % circeVersion)
```
mcovarr: a bit DRYer:

```
libraryDependencies ++= Seq(
  "generic",
  "shapes",
  "refined"
).map(m => "io.circe" %% s"circe-$m" % circeVersion)
```

src/main/scala/wdl4s/cwl/model/CWLType.scala
```diff
@@ -0,0 +1,55 @@
+package broad.cwl.model
```
mcovarr: why the `broad` here?

danbills: whoops!  meant to change the packages.

danbills: will be in line w/ directory i.e. wdl4s.cwl.model

src/main/scala/wdl4s/cwl/model/CWLType.scala
```diff
@@ -0,0 +1,55 @@
...
+
+import enumeratum._
+import mouse.boolean._
+import eu.timepit.refined.api.Refined
+import eu.timepit.refined.string._
+import eu.timepit.refined._
+import eu.timepit.refined.auto._
+
+sealed abstract class CWLType(override val entryName: String) extends EnumEntry
```
mcovarr: We've done all the WDL stuff as `Wdl` rather than `WDL`.

danbills: I can follow that convention, was mirroring the spec spelling

mcovarr: OK.  We should probably discuss this, I think some people in the group may have strong opinions.  😄

src/main/scala/wdl4s/cwl/model/CWLType.scala
```diff
@@ -0,0 +1,55 @@
...
+  path: Option[String],
+  basename: Option[String],
+  dirname: Option[String],
+  nameroot: Option[String],
+  nameext: Option[String],
+  checksum: Option[String],
+  size: Option[Long],
+  secondaryFiles: Array[Either[File, Directory]],
+  format: Option[String],
+  contents: Option[String] //must be
```
mcovarr: cliffhanger comment

mcovarr: or profoundly existential

src/main/scala/wdl4s/cwl/model/CWLVersion.scala
```diff
@@ -0,0 +1,18 @@
...
+  val Draft2Dev2 = Value("draft-3.dev2")
+  val Draft2Dev3 = Value("draft-3.dev3")
+  val Draft2Dev4 = Value("draft-3.dev4")
+  val Draft2Dev5 = Value("draft-3.dev5")
+  val Draft3 = Value("draft-3")
+  val Draft4Dev1 = Value("draft-4.dev1")
+  val Draft4Dev2 = Value("draft-4.dev2")
+  val Draft4Dev3 = Value("draft-4.dev3")
+  val Version1Dev4 = Value("v1.0.dev4")
+  val Version1 = Value("v1.0")
```
mcovarr: 😦

src/main/scala/wdl4s/cwl/model/CWLType.scala
geoffjentry: I feel like I've been summoned

geoffjentry: Oh, yeah. I am a fan of sticking with camel case (or whatever) in the face of acronyms. We do a mix though

mcovarr: cool you have a Gmail filter for "strong opinions"

geoffjentry: I felt a great disturbance in the force

src/main/scala/wdl4s/cwl/model/CWLVersion.scala
mcovarr: It would be scary to actually support all these versions.

src/main/scala/wdl4s/cwl/model/CWLType.scala
```diff
@@ -0,0 +1,55 @@
...
+import eu.timepit.refined.string._
+import eu.timepit.refined._
+import eu.timepit.refined.auto._
+
+sealed abstract class CWLType(override val entryName: String) extends EnumEntry
+
+object CWLType extends Enum[CWLType] {
+  val values = findValues
+
+  case object Null extends CWLType("null")
```
mcovarr: null is a type?

src/main/scala/wdl4s/cwl/model/CWLVersion.scala
danbills: Ah yeah.  We don't support any but the last.  Probably best to just get rid of the other ones.

src/main/scala/wdl4s/cwl/model/CWLType.scala
danbills: Yep http://www.commonwl.org/v1.0/Workflow.html#CWLType

src/main/scala/wdl4s/cwl/model/CWLVersion.scala
mcovarr: Yeah that's the WDL approach, we just don't name our versions.  😄

src/main/scala/wdl4s/cwl/CommandLineTool.scala
```diff
@@ -0,0 +1,97 @@
...
+import eu.timepit.refined.api.Refined
+import eu.timepit.refined.string._
+import shapeless.{:+:, CNil}
+import wdl4s.cwl.CwlVersion._
+import wdl4s.cwl.CwlType._
+
+case class CommandLineTool(
+                            inputs: CommandInputParameter :+: Map[CommandInputParameter#Id, CommandInputParameter#`type`] :+: Map[CommandInputParameter#Id, CommandInputParameter] :+: CNil,
+                            outputs: Array[CommandOutputParameter] :+: Map[CommandOutputParameter#Id, CommandOutputParameter#`type`] :+: Map[CommandOutputParameter#Id, CommandOutputParameter] :+: CNil,
+                            `class`: String, //TODO: must be "CommandLineTool" or ?
```
mcovarr: should this get that
```
String Refined MatchesRegex[W.`"CommandLineTool"`.T]
```
magic or does that not make sense here?

src/main/scala/wdl4s/cwl/ExpressionTool.scala
```diff
@@ -0,0 +1,31 @@
...
+                           hints: Option[Array[String]], //TODO should be Any
+                           label: Option[String],
+                           doc: Option[String],
+                           cwlVersion: Option[CWLVersion]
+)
+
+case class ExpressionToolOutputParameter(
+                                          id: String,
+                                          label: Option[String],
+                                          secondaryFiles: Option[ECMAScriptExpression :+: String :+: Array[ECMAScriptExpression :+: String :+: CNil] :+: CNil], //TODO Use Either if/when it works
```
mcovarr: why the `Either` comment?

build.sbt
```diff
@@ -51,11 +51,28 @@ libraryDependencies ++= {
...
+    "io.circe" %% "circe-yaml" % "0.6.1",
+
+    "eu.timepit" %% "refined"            % "0.8.2",
+    "com.github.benhutchison" %% "mouse" % "0.8",
+
+    "org.scalatest" %% "scalatest" % "3.0.1" % "test",
+    "com.lihaoyi" %% "ammonite-ops" % "1.0.0-RC7" % "test",
+    "org.scalatest" %% "scalatest" % "3.0.1" % "test",
+
     "org.scalatest" %% "scalatest" % "3.0.1" % Test,
```
mcovarr: including scalatest 3 times is the sort of test infrastructure improvement I can get behind 😛

src/main/scala/wdl4s/cwl/CommandLineTool.scala
```diff
@@ -0,0 +1,97 @@
...
+                                  id: Option[String],
+                                  label: Option[String],
+                                  secondaryFiles: Option[Array[ECMAScriptExpression :+: String :+: CNil]],
+                                  format: Option[ECMAScriptExpression :+: Array[String] :+: String :+: CNil], //only valid when type: File
+                                  streamable: Option[Boolean], //only valid when type: File
+                                  doc: Option[String :+: Array[String] :+: CNil],
+                                  inputBinding: Option[CommandLineBinding],
+                                  default: Option[String], //TODO Any type here
+                                  `type`: Option[MyriadInputType]
+
```
mcovarr: This can be a legit `CommandInputParameter` with all of these as `None`s?

src/main/scala/wdl4s/cwl/CommandLineTool.scala
```diff
@@ -0,0 +1,97 @@
...
+) {
+  type Id = String
+  type `type` = MyriadCommandInputType
+}
+
+case class CommandInputRecordSchema(
+  `type`: String Refined MatchesRegex[W.`"record"`.T],
```
mcovarr: is this saying that this can only be constructed with a String literal that's `"record"`?

src/main/scala/wdl4s/cwl/CommandLineTool.scala
mcovarr: or perhaps this shouldn't be a parameter at all if it can only be one possible value?

src/main/scala/wdl4s/cwl/CommandLineTool.scala
```diff
@@ -0,0 +1,97 @@
...
+
+import eu.timepit.refined._
+import eu.timepit.refined.api.Refined
+import eu.timepit.refined.string._
+import shapeless.{:+:, CNil}
+import wdl4s.cwl.CwlVersion._
+import wdl4s.cwl.CwlType._
+
+case class CommandLineTool(
+                            inputs: CommandInputParameter :+: Map[CommandInputParameter#Id, CommandInputParameter#`type`] :+: Map[CommandInputParameter#Id, CommandInputParameter] :+: CNil,
```
mcovarr: I like the way you multilined the coproduct types below, that might work well here too

src/main/scala/wdl4s/cwl/CwlVersion.scala
```diff
@@ -0,0 +1,7 @@
+package wdl4s.cwl
+
+object CwlVersion extends Enumeration {
+  type CWLVersion = Value
```
mcovarr: uppercase escapee

src/main/scala/wdl4s/cwl/model.scala
```diff
@@ -0,0 +1,207 @@
...
+
+case class InputParameter(
+                           id: Option[String], //not really optional but can be specified upstream
+                           label: Option[String],
+                           secondaryFiles: Option[ECMAScriptExpression :+: String :+: Array[ECMAScriptExpression :+: String :+: CNil] :+: CNil],
+                           format: Option[ECMAScriptExpression :+: String :+: Array[String] :+: CNil],
+                           streamable: Option[Boolean],
+                           doc: Option[String :+: Array[String] :+: CNil],
+                           inputBinding: Option[CommandLineBinding],
+                           default: Option[String], //can be of type "Any" which... sucks.
```
mcovarr: `type TrulyMyriad = Any`

src/test/scala/wdl4s/cwl/ParseBigThreeSpec.scala
```diff
@@ -0,0 +1,92 @@
...
+
+  implicit val y = at[Array[WorkflowStep]] {
+    println _
+  }
+}
+
+class ParseBigThreeSpec extends FlatSpec with Matchers {
+  val namespace = "cwl"
+
+  it should "parse 1st tol" in {
```
mcovarr: do tols grow up to be tools?

build.sbt
geoffjentry: Triple redundancy

src/main/scala/wdl4s/cwl/CommandLineTool.scala
```diff
@@ -0,0 +1,97 @@
+package wdl4s.cwl
+
+import eu.timepit.refined._
+import eu.timepit.refined.api.Refined
+import eu.timepit.refined.string._
+import shapeless.{:+:, CNil}
+import wdl4s.cwl.CwlVersion._
+import wdl4s.cwl.CwlType._
+
+case class CommandLineTool(
```
geoffjentry: Since this is our first foray into some of these sigils, can you add comments here on how it is working? Hopefully some day it becomes as second nature as `Option` although "damn I wish there were more docs on our clsses" is a non-infrequent statement I hear at retro and other such things.

src/main/scala/wdl4s/cwl/CommandLineTool.scala
geoffjentry: `id` is "required" but even that's not required by colloquial CWL, just post saladification: http://www.commonwl.org/v1.0/CommandLineTool.html#CommandInputParameter

mcovarr: how would any higher level code sensibly interpret a bag of `None`s?

geoffjentry: Remember that `{}` is valid CWL. As for what *should* happen I dunno. We should probably start becoming regular customers over on the CWL gitter channel - just remember that sometimes it's hard (for outsiders and themselves) to differentiate between "CWL spec" and "What cwltool" does so we need to watch for that

src/main/scala/wdl4s/cwl/CommandLineTool.scala
danbills: Yes, it seems that it should always be "CommandLineTool."  Surprisingly, the spec does not say this, it just says it should be a string.  Hence not enforcing it. 

Generally speaking, in the instances where they are specified "this must be X" it is technically a parameter but enforced to == X by the `MatchesRegex`.  And that should be included in the class to enforce that it is the case.

danbills: I'll clean up the comment to explain the situation

src/main/scala/wdl4s/cwl/ExpressionTool.scala
danbills: If the choice is between 2 types, I'd prefer to use `Either` as it's easier to work with than coproduct.

It's a [long story](https://github.com/circe/circe/issues/672) but we can now go ahead and convert these back to Either.  I just haven't gotten around to doing it and testing it.

So the type within the `Array` becomes `Array[Either[ECMAScriptExpression, String]]`

src/main/scala/wdl4s/cwl/CommandLineTool.scala
danbills: There's an open question as to how to enforce higher-level constraints (e.g. at least one of X, Y,Z should be specified.  I think the solution is either 

1) Figure out how to enforce them with types in this model.  Probably some kind of sealed abstract class with a data type representing each scenario.
2) Do some kind of downstream validation.  This would keep the model more compact but would put the onus on the user to enforce things.

I'd prefer types but it's up for discussion.

src/main/scala/wdl4s/cwl/CommandLineTool.scala
danbills: yes.

Since there are a number of these "type" (and "class") constraints,  we could put them in an Enumeration and each instance would refer to a specific value in that enumeration.  

However because there is only one usage of said value, and there is no real value in the Enumeration itself, might just be better to do it sans structure.

src/main/scala/wdl4s/cwl/CommandLineTool.scala
danbills: TiL "sigil."  I assume you are referring to usage of  `coproduct`?   

I can add them on a package basis, as they would apply generally.

src/main/scala/wdl4s/cwl/model.scala
danbills: Sadly Any is not representable in Json, so a derivation of that type is not possible.  Honestly I'm surprised at such a loosely specified parameter.

What it means in reality is we have to figure out how to support some made-up `Any` type and write a custom parser that can take any of the JSON types.

src/main/scala/wdl4s/cwl/CommandLineTool.scala
geoffjentry: re the quesiton @mcovarr asked, I went to the gitter channel and they said: "It would still be available for referencing from the arguments section, for example
That pattern (just a list of input names and their types, and using the arguments section for composing the command line) is likely to become the default way: common-workflow-language/common-workflow-language#373"

re the options @danbills gave, in such questions I'm almost always going to side with types :)

src/main/scala/wdl4s/cwl/ExpressionTool.scala
mcovarr: hmm I'd probably vote for consistency if we know we're going to have situations with more than two types.

src/main/scala/wdl4s/cwl/CommandLineTool.scala
geoffjentry: Yeah :) In general though I was just suggesting adding some comments on how this whole approach

build.sbt
```diff
@@ -64,6 +78,8 @@ libraryDependencies ++= {
 // https://github.com/scala/pickling/issues/10
 scalacOptions ++= Seq("-deprecation", "-unchecked", "-feature", "-Xmax-classfile-name", "200")
 
+addCompilerPlugin("org.scalamacros" % "paradise" % "2.1.0" cross CrossVersion.full)
```
rtitle: I'm curious why macro-paradise is needed -- is it a requirement for circe-generic?

src/main/scala/wdl4s/cwl/package.scala
```diff
@@ -0,0 +1,172 @@
...
+  import CwlType._
+  import CwlVersion._
+  import ScatterMethod._
+
+
+  implicit val cwlTypeDecoder = Decoder.enumDecoder(CwlType)
+  implicit val cwlVersionDecoder = Decoder.enumDecoder(CwlVersion)
+  implicit val scatterMethodDecoder = Decoder.enumDecoder(ScatterMethod)
+
+  implicit private val workflowDecoder: Decoder[Workflow] = deriveDecoder[Workflow]
```
rtitle: I'm impressed, circe automatic derivation "just works" with shapeless coproducts & refined types?

danbills: yeah but at the cost of compilation time.  this file takes ~ 30s !

build.sbt
danbills: Yes it is

src/main/scala/wdl4s/cwl/package.scala
rtitle: https://github.com/milessabin/shapeless/blob/master/core/src/main/scala/shapeless/test/compiletime.scala

:)

danbills:
Am realizing that if anybody wanted to write CWL, defaulting all these `Option`s to None would be an absolute necessity.

cjllanwarne:
TOL: Would it save us headaches if we merge this into `cromwomification` first rather than `develop?`

src/main/scala/wdl4s/cwl/CommandLineTool.scala
curoli: @geoffjentry, @danbills: Just food for thought: I love expressing things with types, but I can't imagine you can express nearly all constraints of the CWL specs as types. So there is going to be extensive downstream validation no matter what. I'm also weary of using Shapeless extensively, or using things such as type co-products, since the maintenance burden is quite high: developers will struggle to get it right, the compiler will take an enormous time, and error messages will be cryptic. I've also seen code that compiled fine but IntelliJ would insist it was illegal.
But that's just food for thought, feel free to go ahead, and if it doesn't work out, we can always switch to a simpler type system.

danbills:
@cjllanwarne This is independent of cromwom, I don't see any headaches coming.

src/main/scala/wdl4s/cwl/CommandLineTool.scala
danbills: @curoli duly noted

src/main/scala/wdl4s/cwl/Workflow.scala
```diff
@@ -0,0 +1,3 @@
+package wdl4s.cwl
```
geoffjentry: wht's up w/ this file?

cjllanwarne:
@danbills - the only future headache would be if we want to get a head start on https://github.com/broadinstitute/cromwell/issues/2383

src/main/scala/wdl4s/cwl/Workflow.scala
danbills: It escaped the `git rm` cleaver. brb

danbills:
@cjllanwarne per our discussion a rebase would have to happen regardless and this code should not hinder that

curoli:
I think we should add type aliases for all type coproducts used. For example, in 
```
case class CommandLineTool(
  inputs:
    CommandInputParameter :+:
    Map[CommandInputParameter#Id, CommandInputParameter#`type`] :+:
    Map[CommandInputParameter#Id, CommandInputParameter] :+:
    CNil,
...)
```
To assign inputs, we need to create
```
Coproduct[CommandInputParameter :+:
                  Map[CommandInputParameter#Id, CommandInputParameter#`type`] :+:
                  Map[CommandInputParameter#Id, CommandInputParameter] :+:
                  CNil](...)
```
There should be some type alias for
```
CommandInputParameter :+: 
Map[CommandInputParameter#Id, CommandInputParameter#`type`] :+:
Map[CommandInputParameter#Id, CommandInputParameter] :+: 
CNil
```

danbills:
@curoli That's fine but like I pointed out in the issue there is an `inject` feature for building coproducts that will accept any value and build the correctly typed coproduct for you.

This would save us from having numerous types named only for the purpose of reducing this boilerplate.

The catch is that this feature is brand new so is not part of a published shapeless release.

curoli:
@danbills Ok, so you're saying we won't need those type aliases in the future. But what we are going to do now? How is a class like CommandLineTool instantiated with circe? Does circe have some built-in support for shapeless type coproducts? Somehow, I can't find where we instantiate objects to be assigned to references with coproduct types.

danbills:
@curoli Let's see if we can publish shapeless snapshot to our own artifactory and use that.  I think the type aliases will be noisy and ultimately unnecessary.

The reason you are not seeing instantiations anywhere and coproduct assignments is because circe is doing it for us when it parses the YAML input.

I'll take a look tomorrow when I get in and if successful I will push something simple to demonstrate

curoli:
(Let me know if we should move this conversation elsewhere, if it is not sufficiently related to this PR)

If you think using a shapeless snapshot is the best solution, that's fine with me. Alternatively, I suppose we could write our own convenience methods as a stop-gap. Is [shapeless/syntax/inject.scala](https://github.com/milessabin/shapeless/blob/master/core/src/main/scala/shapeless/syntax/inject.scala) the only code needed?

That's actually really awesome that circe can auto-decode objects with coproducts. I'm assuming it should be able to auto-encode them, too, then.

Any examples would be greatly appreciated.

danbills:
@curoli I pushed some type aliases and some code to make it less gross.  Let's discuss more over at your ticket for exporting CWL as it's not related to parsing per se

geoffjentry:
👍

curoli:
@danbills Thanks, sounds very useful. Yeah, let's discuss at my branch. 

I successfully created a CWL CommandLineTool (very verbose, but straight-forward). Now I'm trying to convince circe to encode it.

src/main/scala/wdl4s/cwl/CwlType.scala
```diff
@@ -0,0 +1,51 @@
...
+  checksum: Option[String],
+  size: Option[Long],
+  secondaryFiles: Array[File :+: Directory :+: CNil],
+  format: Option[String],
+  contents: Option[String])
+
+object File {
+  def validateContents: File => Either[String, Unit] =  //this is a sample validation.  There are about a billion of them, not sure
+    f =>
+      (!(f.location.isEmpty && f.path.isEmpty && f.contents.isEmpty)).
```
mcovarr: is it cool if more than one is not empty?

src/main/scala/wdl4s/cwl/CommandLineTool.scala
```diff
@@ -0,0 +1,79 @@
...
+                                      CommandInputArraySchema :+:
+                                      String :+:
+                                      Array[
+                                        CwlType :+:
+                                          CommandInputRecordSchema :+:
+                                          CommandInputEnumSchema :+:
+                                          CommandInputArraySchema :+:
+                                          String :+:
+                                          CNil
+                                        ] :+:
```
mcovarr: the indent on these coproducts seems slightly goofed up

mcovarr:
:+1: not sure I understand it 100% but looking forward to learning it 😄

src/main/scala/wdl4s/cwl/CwlType.scala
danbills: the validation of Files is [painful](http://www.commonwl.org/v1.0/Workflow.html#File)  I thought I could get it started but I am going to create a ticket as it is pretty tricky and this is nowhere near complete.  

In fact I'm just going to delete this as it's the only piece of downstream validation there is.

PR #121, opened 2017-06-26
Add trailing newline in tsv serialization

mcovarr:
Changes to add trailing newlines in TSV serialization, for the benefit of write_lines and other functions.  There's a corresponding Centaur branch which allows for the removal of some [hackery](https://github.com/broadinstitute/centaur/commit/c770edbd2c98ba12d46051164232ae8c5e60012b).

### DIFF
--- a/src/main/scala/wdl4s/values/WdlArray.scala
+++ b/src/main/scala/wdl4s/values/WdlArray.scala
@@ -58,12 +58,13 @@ sealed abstract case class WdlArray(wdlType: WdlArrayType, value: Seq[WdlValue])
 
   def tsvSerialize: Try[String] = {
     wdlType.memberType match {
-      case t: WdlPrimitiveType => Success(value.map(_.valueString).mkString("\n"))
+      case t: WdlPrimitiveType => Success(value.map(_.valueString).mkString(start = "", sep = "\n", end = "\n"))
       case WdlObjectType => WdlObject.tsvSerializeArray(value map { _.asInstanceOf[WdlObject] })
       case WdlArrayType(t: WdlPrimitiveType) =>
         val tsvString = value.collect({ case a: WdlArray => a }) map { a =>
-          a.value.collect({ case p: WdlPrimitive => p }).map(_.valueString).mkString("\t")
-        } mkString "\n"
+          a.value.collect({ case p: WdlPrimitive => p.valueString }).mkString(start = "", sep = "\t", end = "\n")
+        } mkString
+
         Success(tsvString)
       case _ => Failure(new UnsupportedOperationException(s"Cannot TSV serialize a ${this.wdlType.toWdlString} (valid types are Array[Primitive], Array[Array[Primitive]], or Array[Object])"))
     }
--- a/src/main/scala/wdl4s/values/WdlObject.scala
+++ b/src/main/scala/wdl4s/values/WdlObject.scala
@@ -2,7 +2,7 @@ package wdl4s.values
 
 import wdl4s.types._
 import wdl4s.util.FileUtil
-import wdl4s.{Call, TaskCall, TsvSerializable}
+import wdl4s.{Call, TsvSerializable}
 
 import scala.util.{Failure, Success, Try}
 
@@ -53,7 +53,7 @@ object WdlObject {
         val attributesLine = attributes.mkString("\t")
         val valuesLines = objects map { obj =>
           attributes map { obj.value(_).valueString } mkString "\t"
-        } mkString "\n"
+        } mkString(start = "", sep = "\n", end = "\n")
 
         Success(s"$attributesLine\n$valuesLines")
       case _ => Failure(new UnsupportedOperationException("Could not serialize array: Objects in the array have different attributes."))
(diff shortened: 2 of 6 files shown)


src/main/scala/wdl4s/values/WdlArray.scala
```diff
@@ -58,12 +58,12 @@ sealed abstract case class WdlArray(wdlType: WdlArrayType, value: Seq[WdlValue])
 
   def tsvSerialize: Try[String] = {
     wdlType.memberType match {
-      case t: WdlPrimitiveType => Success(value.map(_.valueString).mkString("\n"))
+      case t: WdlPrimitiveType => Success(value.map(_.valueString + "\n").mkString(""))
```
cjllanwarne: `mkString` actually has a "suffix" parameter you can use for this.
Or, if you leave it like this, you don't need arguments at all, ie just `.mkString`

src/main/scala/wdl4s/values/WdlArray.scala
```diff
@@ -58,12 +58,12 @@ sealed abstract case class WdlArray(wdlType: WdlArrayType, value: Seq[WdlValue])
...
       case WdlObjectType => WdlObject.tsvSerializeArray(value map { _.asInstanceOf[WdlObject] })
```
cjllanwarne: Presumably we should also enhance this to append a `\n`?

src/main/scala/wdl4s/values/WdlArray.scala
```diff
@@ -58,12 +58,12 @@ sealed abstract case class WdlArray(wdlType: WdlArrayType, value: Seq[WdlValue])
...
       case WdlArrayType(t: WdlPrimitiveType) =>
         val tsvString = value.collect({ case a: WdlArray => a }) map { a =>
-          a.value.collect({ case p: WdlPrimitive => p }).map(_.valueString).mkString("\t")
-        } mkString "\n"
+          a.value.collect({ case p: WdlPrimitive => p }).map(_.valueString).mkString("\t") + "\n"
+        } mkString ""
```
cjllanwarne: Same here, could use the parameterless version

cjllanwarne:
Looks good, only minor nitpickery. 👍

src/main/scala/wdl4s/values/WdlArray.scala
mcovarr: I actually had to add a newline so as not to confuse the parsing, but ok

src/main/scala/wdl4s/values/WdlArray.scala
mcovarr: Yes, yes we should.  And we should have a test for that and the other cases here.

geoffjentry:
:+1:

Out of curiosity, can you think of any legitimate reason why someone wouldn't want this? I have a tiny backwards compat concern but I can't think of why it'd matter.

mcovarr:
@geoffjentry backwards compat is my only concern, I think it was wrong before.

geoffjentry:
@mcovarr right - i was questioning if that even mattered, ie could this actually screw someone up? I'd think that any reasonable TSV parser wouldn't mind.

mcovarr:
Well this affects `write_lines` and friends too.  So if anything out there is doing arithmetic based on the number of lines emitted by `write_lines`, then this might cause breakages.

mcovarr:
@geoffjentry And thanks to @cjllanwarne's eagle-eyed observations, I have learned that the current wdl4s TSV parser would not meet your criteria for "reasonable".  😆   More changes forthcoming...

src/test/scala/wdl4s/values/WdlObjectSpec.scala
```diff
@@ -4,11 +4,11 @@ import wdl4s.types.{WdlArrayType, WdlObjectType}
...
 
 class WdlObjectSpec extends FlatSpec with Matchers with TryValues {
-  val correctTSV = "one\ttwo\tthree\tfour\none\tfour\tnine\tsixteen"
+  val correctTSV = "one\ttwo\tthree\tfour\none\tfour\tnine\tsixteen\n"
   val emptyTSV = ""
-  val oneRowTSV = "one\ttwo\tthree\tfour"
-  val nonHomogeneousTS = "onet\ttwo\tthree\none\ttwo"
-  val arrayTSV = correctTSV + "\none\teight\ttwentyseven\tsixtyfour"
+  val oneRowTSV = "one\ttwo\tthree\tfour\n"
+  val nonHomogeneousTS = "one\ttwo\tthree\none\ttwo\n"
```
mcovarr: preexisting wackiness, didn't make sense to me either.  🙁

src/main/scala/wdl4s/values/WdlObject.scala
```diff
@@ -78,7 +79,7 @@ case class WdlObject(value: Map[String, WdlValue]) extends WdlValue with WdlObje
       case _ => throw new UnsupportedOperationException("Can only TSV serialize an Object with Primitive values.")
     }
 
-    s"$keysLine\n${values.mkString("\t")}"
+    keysLine + "\n" + values.mkString("\t") + "\n"
```
mcovarr: oh you are crafty

PR #126, opened 2017-06-29
Have explicit TaskNode and WorkflowNode

Horneth:
I think this will be better to make things more typesafe in Cromwell e.g. where we know that we want a TaskCall and not a a WorkflowCall (to give to a backend for example).

### DIFF
--- a/src/main/scala/wdl4s/wom/graph/CallNode.scala
+++ b/src/main/scala/wdl4s/wom/graph/CallNode.scala
@@ -8,14 +8,23 @@ import wdl4s.wom.graph.GraphNodePort.{ConnectedInputPort, DeclarationOutputPort,
 import scala.language.postfixOps
 import cats.implicits._
 
-final case class CallNode private(name: String, callable: Callable, inputPorts: Set[GraphNodePort.InputPort]) extends GraphNode {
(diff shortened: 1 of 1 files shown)


geoffjentry:
👍

src/main/scala/wdl4s/wom/graph/CallNode.scala
```diff
@@ -8,14 +8,23 @@ import wdl4s.wom.graph.GraphNodePort.{ConnectedInputPort, DeclarationOutputPort,
...
+sealed abstract class CallNode extends GraphNode {
+  def name: String
+  def callable: Callable
+  def callType: String
+}
+
+final case class TaskCallNode private(name: String, callable: TaskDefinition, inputPorts: Set[GraphNodePort.InputPort]) extends CallNode {
+  val callType: String = "task"
+  override val outputPorts: Set[GraphNodePort.OutputPort] = {
+    callable.outputs.map(o => DeclarationOutputPort(o.name, o.womType, this))
```
mcovarr: nice, this was bugging me in the baseline version.  😄

mcovarr:
:+1:

PR #135, opened 2017-07-10
Surgical prep prior to subprojecting.

mcovarr:
I'm hoping to end up with an sbt structure similar to Cromwell's, so I started with a Cromwellesque set of compiler flags and the associated cleanup.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -5,9 +5,9 @@ name := "wdl4s"
 
 organization := "org.broadinstitute"
 
-scalaVersion := "2.12.1"
+scalaVersion := "2.12.2"
 
-crossScalaVersions := Seq("2.11.8", "2.12.1")
+crossScalaVersions := Seq("2.11.8", "2.12.2")
 
 lazy val versionSettings = Seq(
   // Upcoming release, or current if we're on the master branch
@@ -46,7 +46,9 @@ libraryDependencies ++= {
      */
     "org.typelevel" %% "cats" % "0.9.0"
       exclude("org.typelevel", "cats-laws_2.11")
-      exclude("org.typelevel", "cats-kernel-laws_2.11"),
+      exclude("org.typelevel", "cats-laws_2.12")
+      exclude("org.typelevel", "cats-kernel-laws_2.11")
+      exclude("org.typelevel", "cats-kernel-laws_2.12"),
     "commons-codec" % "commons-codec" % "1.10",
     "commons-io" % "commons-io" % "2.5",
     "org.apache.commons" % "commons-lang3" % "3.4",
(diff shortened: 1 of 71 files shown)


geoffjentry:
Too soon for the surgical jokes ;)

build.sbt
```diff
@@ -5,9 +5,7 @@ name := "wdl4s"
 
 organization := "org.broadinstitute"
 
-scalaVersion := "2.12.1"
-
-crossScalaVersions := Seq("2.11.8", "2.12.1")
```
geoffjentry: I'm pretty sure there are other projects out there consuming wdl4s

geoffjentry: Like [Rawls](https://github.com/broadinstitute/rawls/blob/develop/project/Dependencies.scala#L67)

mcovarr: boo

mcovarr:
😷

src/main/scala/wdl4s/wdl.scala
```diff
@@ -0,0 +1,20 @@
+package wdl4s
+
+import wdl4s.wdl.exception.OutputVariableLookupException
+import wdl4s.wdl.values.WdlValue
+
+import scala.util.{Failure, Try}
+
```
geoffjentry: Some of these look like they'd be used outside of `wdl`, is the idea that they'd move elsewhere?

src/main/scala/wdl4s/wdl/WdlNamespace.scala
```diff
@@ -494,15 +479,13 @@ object WdlNamespace {
...
-    /**
-      * Ensures that the lhs corresponds to a call and the rhs corresponds to one of its outputs. We're only checking
-      * top level MemberAccess ASTs because the sub-ASTs don't make sense w/o the context of the parent. For example
-      * if we have "input: var=ns.ns1.my_task" it does not make sense to validate "ns1.my_task" by itself as it only
-      * makes sense to validate that "ns.ns1.my_task" as a whole is coherent
-      *
-      * Run for its side effect (i.e. Exception) but we were previously using a Try and immediately calling .get on it
-      * so it's the same thing
-      */
+    // Ensures that the lhs corresponds to a call and the rhs corresponds to one of its outputs. We're only checking
```
geoffjentry: FWIW I prefer the `/* ... */` style for multiline comments. Realizing that that's a *third* option beyond what's presented in the diff I'll just leave that as a TOL :)

src/main/scala/wdl4s/wdl/WdlNamespace.scala
```diff
@@ -494,15 +479,13 @@ object WdlNamespace {
       }
     }
 
-    /**
-      * Ensures that the lhs corresponds to a call and the rhs corresponds to one of its outputs. We're only checking
-      * top level MemberAccess ASTs because the sub-ASTs don't make sense w/o the context of the parent. For example
-      * if we have "input: var=ns.ns1.my_task" it does not make sense to validate "ns1.my_task" by itself as it only
-      * makes sense to validate that "ns.ns1.my_task" as a whole is coherent
-      *
-      * Run for its side effect (i.e. Exception) but we were previously using a Try and immediately calling .get on it
```
geoffjentry: I bet I know who wrote this comment originally

src/main/scala/wdl4s/wdl.scala
mcovarr: oh yeah all this stuff is going to move in a sequel PR

src/main/scala/wdl4s/wdl/expression/WdlStandardLibraryFunctions.scala
```diff
@@ -77,8 +76,8 @@ trait WdlStandardLibraryFunctions extends WdlFunctions[WdlValue] {
   def basename(params: Seq[Try[WdlValue]]): Try[WdlString] = {
 
     val arguments: Try[(WdlValue, WdlValue)] = params.toList match {
-      case Success(f) :: Nil => Success(f, WdlString(""))
-      case Success(f) :: Success(s) :: Nil => Success(f, s)
+      case Success(f) :: Nil => Success((f, WdlString("")))
```
geoffjentry: curiosity: was this running afoul of the eta expansion warnings?

src/main/scala/wdl4s/wdl/WdlNamespace.scala
mcovarr: That's fine, it just can't be scaladoc or we get the "unmoored" errors.

src/main/scala/wdl4s/wdl/expression/WdlStandardLibraryFunctions.scala
mcovarr: If I remember what eta expansion is, no; this was running afoul of relying on auto-tupling.

src/main/scala/wdl4s/wdl/util/StringUtil.scala
```diff
@@ -47,14 +47,14 @@ object StringUtil {
     @tailrec
```
geoffjentry: I suspect these could be folds, but since that condition isn't the one `wdl4s` is going under the knife for I'll just leave that TOL there

src/main/scala/wdl4s/wdl/expression/WdlStandardLibraryFunctions.scala
geoffjentry: Oh right. I was picturing the right situation but the wrong error message

src/main/scala/wdl4s/wdl/WdlNamespace.scala
geoffjentry: Yeah. I don't like using the `/**` outside of actual scaladoc places. While I know I wrote some of that text I don't think I was the one who caused the extra star to happen.

geoffjentry:
:+1: top notch slicin' and dicin'

build.sbt
```diff
@@ -76,11 +74,53 @@ libraryDependencies ++= Seq(
...
+  "-Ywarn-dead-code",
+  "-Ywarn-numeric-widen",
+  "-Ywarn-value-discard",
+  "-Ywarn-inaccessible",
+  "-Ywarn-unused:implicits",
+  "-Ywarn-unused:imports",
+  "-Ywarn-unused:privates",
+  "-Ywarn-unused:locals",
+  "-Xfatal-warnings",
+  "-Ywarn-unused:patvars"
```
kshakir: Just like in the cromwell PR, group the warn-unused.

src/test/scala/wdl4s/wdl/values/WdlValueSpec.scala
```diff
@@ -50,7 +50,7 @@ class WdlValueSpec extends FlatSpec with Matchers {
     (WdlFloat(Double.MinValue), "-1.7976931348623157E308", { d: Double => d.isNegInfinity }))
 
   forAll(wdlFloatSpecials) { (wdlValue, rawString, validateFloat) =>
-    it should s"convert a special ${wdlValue.typeName} to/from raw string '$rawString'" in {
+    it should s"convert a special ${wdlValue.typeName} to/from raw string '$rawString'" ignore {
```
kshakir: Why is this ignored, and when is it expected to get un-ignored?

kshakir:
👍  Once comments addressed.

src/test/scala/wdl4s/wdl/values/WdlValueSpec.scala
mcovarr: Great catch!  I meant to drop a comment here and completely forgot.

The problem here was the old behavior at `WdlFloatType.fromWdlString` above.  This method takes a string and used to invoke `toFloat` rather than `toDouble`.  This is curious considering that the internal data type in a `WdlFloat` is in fact a `double`, and the compiler was silently widening the `float` into a `double` after this conversion anyway (I had to make this change because the new flags disapprove of these silent widenings).

These tests were validating that any values that fell between the limits of a `float` and a `double` were truncated by the `toFloat`.  I have no idea why that truncation behavior wasn't fixed rather than writing tests to validate that was happening.  I'd probably want to replace these with tests that validated truncation outside the limits of doubles, or just delete them altogether.

mcovarr:
@geoffjentry @kshakir A re-review of just `build.sbt` would be appreciated.  I made some changes to reenable support for cross-compiling/publishing, as well as making `sbt console` usable.

build.sbt
```diff
@@ -76,11 +78,76 @@ libraryDependencies ++= Seq(
...
 
+scalacOptions ++= (CrossVersion.partialVersion(scalaVersion.value) match {
+  case Some((2, 12)) =>
+    // The default scalacOptions includes console-hostile options.  These options are overridden specifically below
+    // for the `console` target.
+    baseOptions ++ warningOptions ++ consoleHostileOptions
+  case Some((2, 11)) =>
+    // Scala 2.11 takes a simplified set of options
+    baseOptions
+  case _ => ??? // wdl4s has not been tested with earlier versions of Scala.
```
geoffjentry: NotImplementedException w/ this message

geoffjentry:
:100: / :100:

src/test/scala/wdl4s/wdl/values/WdlValueSpec.scala
mcovarr: I think I'm just going to delete these tests unless there are any objections.  The tested behavior has been eliminated and I think the world is a better place for that.

PR #137, opened 2017-07-12
Subprojecting part II: CWL and WOM, with WOM currently including all WDL.

mcovarr:


geoffjentry:
Since this is more structural than codeual, could you give a bit of a description here of how things are laid out now and how it's all intended to fit together? It'd be easier to go through the PR armed with that knowledge.

mcovarr:
This creates subprojects `wom`, `wdl`, and `cwl`.  `wdl` and `cwl` should contain only the implementation-dependent bits and depend on `wom`.  `wom` should contain the implementation-agnostic bits and depend on no other subproject.  

In this PR `cwl` conforms to this plan, but `wdl` and `wom` are a mess.  Everything is dumped into `wom` since we're using WDL types and values at the WOM level, but those types and values currently spider out into WDL parsing in ways that are not trivial to separate.  Separating WDL stuff into `wdl` will be the subject of PR III.

geoffjentry:
👍

danbills:
👍 
props for organization

PR #147, opened 2017-07-19
FileEvaluator handles: two parameter functions and function calls

cjllanwarne:
To enable output blocks like:
```
output {
  # Function call in a function call:
  Int x = round(size("myfile.txt"))
  # Two parameter function call:
  Float y = size("myFile.txt", "GB")
}
```

### DIFF
--- a/wom/src/main/scala/wdl4s/wdl/WdlExpression.scala
+++ b/wom/src/main/scala/wdl4s/wdl/WdlExpression.scala
@@ -33,11 +33,10 @@ object WdlExpression {
     def isMapLiteral: Boolean = ast.getName == "MapLiteral"
     def isObjectLiteral: Boolean = ast.getName == "ObjectLiteral"
     def isArrayOrMapLookup: Boolean = ast.getName == "ArrayOrMapLookup"
-    def params = ast.getAttribute("params").asInstanceOf[AstList].asScala.toVector
+    def params: Vector[AstNode] = Option(ast.getAttribute("params")).map(_.asInstanceOf[AstList].asScala.toVector).getOrElse(Vector.empty)
     def name = ast.getAttribute("name").asInstanceOf[Terminal].getSourceString
-    def isFunctionCallWithOneParameter = ast.isFunctionCall && ast.params.size == 1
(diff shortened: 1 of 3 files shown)


wom/src/main/scala/wdl4s/wdl/WdlExpression.scala
```diff
@@ -33,11 +33,10 @@ object WdlExpression {
     def isMapLiteral: Boolean = ast.getName == "MapLiteral"
     def isObjectLiteral: Boolean = ast.getName == "ObjectLiteral"
     def isArrayOrMapLookup: Boolean = ast.getName == "ArrayOrMapLookup"
-    def params = ast.getAttribute("params").asInstanceOf[AstList].asScala.toVector
+    def params: Vector[AstNode] = Option(ast.getAttribute("params")).map(_.asInstanceOf[AstList].asScala.toVector).getOrElse(Vector.empty)
     def name = ast.getAttribute("name").asInstanceOf[Terminal].getSourceString
-    def isFunctionCallWithOneParameter = ast.isFunctionCall && ast.params.size == 1
-    def isFunctionCallWithOneFileParameter = isFunctionCallWithOneParameter && WdlFunctionsWithSingleFileParameter.contains(ast.functionName)
-    def isGlobFunctionCall = isFunctionCallWithOneParameter && "glob".equals(ast.functionName)
+    def isFunctionCallWithFirstParameterBeingFile = ast.isFunctionCall && ast.params.nonEmpty && WdlFunctionsWithFirstParameterBeingFile.contains(ast.functionName)
```
geoffjentry: I find this name awkward but have no better suggestion. If no one else has a problem with this ignore my comment

geoffjentry:
👍

wom/src/main/scala/wdl4s/wdl/WdlExpression.scala
cjllanwarne: Agreed... and agreed.

mcovarr: res ipsum loquitur and all, but yeah it does kind of go on for a while.  🙂   I think it's fine.

mcovarr:
TIL those expressions wouldn't have worked before.  😦   Nice!  :+1:

PR #150, opened 2017-07-25
Add creating and encoding workflow and command line tool with requirements

curoli:
Creating and encoding as CWL a CWL workflow object and a CWL command line tool object with a requirement.


curoli:
This fixes #125.

geoffjentry:
@curoli Why are you pulling in old commits?

cwl/src/main/scala/wdl4s/cwl/package.scala
```diff
@@ -91,4 +91,19 @@ package object cwl extends TypeAliases {
     import wdl4s.cwl.Implicits.enumerationEncoder
     commandLineTool.asJson
   }
+
```
geoffjentry: Can we maybe pull the encoder/decoders (these plus the older ones) into a separate object to give it a more clear namespace than the random grab bag of fun a package object provides?

curoli:
@geoffjentry You mean the two commits from @danbills? Well, it was useful to have them, and I expected to merge after him, which I did. I suppose they appear here because @danbills squashed them away before merging. What to do in such cases? In my other projects, we didn't squash, so they would not appear.

cwl/src/main/scala/wdl4s/cwl/package.scala
curoli: I have no objection to that. In my previous PR, I put them in an object CwlEncoders and was asked to put them in the package instead. Where should they go?

cwl/src/main/scala/wdl4s/cwl/package.scala
```diff
@@ -91,4 +91,19 @@ package object cwl extends TypeAliases {
...
+  def encodeCwlWorkflow: Workflow => Json = { workflow =>
```
danbills: ToL I kinda like it this way, which is not surprising because I wrote it.

curoli: I wrote it to look like the decodeCwl method. But actually, a simple one-argument method may be nicer.

cwl/src/test/scala/wdl4s/cwl/ExportCwlSamplesSpec.scala
```diff
@@ -28,27 +39,166 @@ class ExportCwlSamplesSpec extends FlatSpec with Matchers {
...
+        |  "baseCommand" : "echo"
+        |}""".stripMargin
+    assertCorrectJson(tool, expectedToolJsonString)
+  }
+
+  it should "encode sample CWL workflow" in {
+    val workflow = Workflow(
+      cwlVersion = Option(CwlVersion.Version1),
+//      `class` = "Workflow",
+      inputs = Coproduct[WorkflowInput](
```
danbills: Agree the `Coproduct` is noisy.  In some cases the type argument isn't necessary but I can't figure out when the compiler needs it and when it doesn't.

In any case I haven't figured out how to implicitly create all these `Coproduct`s.

cwl/src/main/scala/wdl4s/cwl/package.scala
danbills: I should explain that I like it because 

 * when you are composing functions you don't have to partially apply them. 
 * you don't have to name the argument, i.e. you get a nice underscore to use.
 * It's easier to read IMO

curoli: But you can always lift a one-argument method into a Function1, e.g.:
```
scala> def m(i: Int): Int = 2*i
m: (i: Int)Int

scala> Seq(1, 2, 3).map(m)
res0: Seq[Int] = List(2, 4, 6)
```

geoffjentry:
@curoli I think if you were to rebase it'd fix the commits

cwl/src/main/scala/wdl4s/cwl/package.scala
geoffjentry: I have a strong aversion to package objects starting to accumulate stuff. 

Unless there's a good reason not to (and others, including whomever made the original request can chime in) I'd prefer to see these in a separate object. I don't really care what the name is.

cwl/src/main/scala/wdl4s/cwl/package.scala
danbills: go ahead w/ one-arg function, we're delving into nitpickiness and it's not worth it.

cwl/src/main/scala/wdl4s/cwl/model.scala
```diff
@@ -207,12 +208,15 @@ case class Dirent(
   * @param envDef
   */
 case class EnvVarRequirement(
-  `class`: Witness.`"EnvVarRequirement"`.T,
-  envDef:
-    Array[EnvironmentDef] :+:
-    Map[EnvironmentDef#EnvName, EnvironmentDef#EnvValue] :+:
-    Map[EnvironmentDef#EnvName, EnvironmentDef] :+:
-    CNil)
+                              `class`: Witness.`"EnvVarRequirement"`.T,
```
mcovarr: I'd guess because you're not initializing to a default value here and you are in the other spots

cwl/src/main/scala/wdl4s/cwl/package.scala
danbills: I like having the package `import wdl4s.cwl._` but it's just a nice to have.  If @geoffjentry feels strongly otherwise let's do it somewhere else.

cwl/src/main/scala/wdl4s/cwl/model.scala
curoli: Yeah, most likely.

cwl/src/main/scala/wdl4s/cwl/package.scala
curoli: I moved them now to cwl.CwlCodecs.

geoffjentry: @danbills For me it is mainly that I've seen too many package objects become giant bags of random stuff. Once that point hits it's impossible to say "don't put that here" because everything else is already in there.

cwl/src/main/scala/wdl4s/cwl/CwlCodecs.scala
```diff
@@ -0,0 +1,59 @@
...
+import eu.timepit.refined.api.Refined
+import eu.timepit.refined.string._
+import eu.timepit.refined._
+import io.circe.refined._
+import io.circe.literal._
+
+/**
+  * wdl4s
+  * Created by oruebenacker on 25.07.17.
+  */
```
mcovarr: remove plz

curoli: Removed (note to self: switch off header generation in IntelliJ)

cwl/src/main/scala/wdl4s/cwl/Cwl.scala
```diff
@@ -14,11 +14,15 @@ sealed trait Cwl {
...
 case class Workflow(
   cwlVersion: Option[CwlVersion] = None,
-  `class`: Witness.`"Workflow"`.T = "Workflow".narrow,
+  `class` : Workflow.`class`.type = Workflow.`class`,
   inputs: WorkflowInput,
   outputs: WorkflowOutput,
   steps: WorkflowSteps) extends Cwl
 
+object Workflow {
+  val `class` : Witness.`"Workflow"`.T = "Workflow".narrow
```
danbills: Please inline this to Workflow's default.

cwl/src/main/scala/wdl4s/cwl/Cwl.scala
```diff
@@ -42,7 +46,7 @@ case class Workflow(
 case class CommandLineTool(
                             inputs: Inputs = Coproduct[Inputs](Array.empty[CommandInputParameter]),
                             outputs: Outputs = Coproduct[Outputs](Array.empty[CommandOutputParameter]),
-                            `class`: W.`"CommandLineTool"`.T = "CommandLineTool".narrow,
+                            `class`: CommandLineTool.`class`.type = CommandLineTool.`class`,
```
danbills: Inline please as above

cwl/src/test/scala/wdl4s/cwl/ExportCwlSamplesSpec.scala
```diff
@@ -28,27 +39,130 @@ class ExportCwlSamplesSpec extends FlatSpec with Matchers {
...
+            id = None,
+            `type` = Option(Coproduct[MyriadOutputType](CwlType.File)),
+            outputSource = Option(Coproduct[WorkflowOutputParameter#OutputSource]("compile/classfile"))
+          )
+        )
+      ),
+      steps = Coproduct[WorkflowSteps](
+        Map(
+          "untar" -> WorkflowStep(
+            id = None,
```
danbills: For all these places that you're using `None`, could you please make this the default in the class constructor parameter?  Then please omit them in this test.

cwl/src/test/scala/wdl4s/cwl/ExportCwlSamplesSpec.scala
```diff
@@ -28,27 +39,130 @@ class ExportCwlSamplesSpec extends FlatSpec with Matchers {
...
+        |""".stripMargin
+    assertCorrectJson(workflow, expectedWorkflowJsonString)
+  }
+
+  it should "encode sample CWL env" in {
+    val tool = CommandLineTool(
+      cwlVersion = Option(CwlVersion.Version1),
+      baseCommand = Option(Coproduct[BaseCommand]("env")),
+      requirements = Option(Array(Coproduct[Requirement](EnvVarRequirement(
+        `class` = "EnvVarRequirement",
```
danbills: Could you please make this default and omit here?

cwl/src/test/scala/wdl4s/cwl/ExportCwlSamplesSpec.scala
```diff
@@ -28,27 +39,130 @@ class ExportCwlSamplesSpec extends FlatSpec with Matchers {
...
+        |    out:
+        |    - classfile
+        |    run: arguments.cwl
+        |""".stripMargin
+    assertCorrectJson(workflow, expectedWorkflowJsonString)
+  }
+
+  it should "encode sample CWL env" in {
+    val tool = CommandLineTool(
+      cwlVersion = Option(CwlVersion.Version1),
```
danbills: We should probably make this default too

cwl/src/test/scala/wdl4s/cwl/ExportCwlSamplesSpec.scala
```diff
@@ -1,12 +1,23 @@
...
+import wdl4s.cwl.WorkflowStep.{Inputs, Run}
 
 class ExportCwlSamplesSpec extends FlatSpec with Matchers {
 
-  it should "export 1st tool" in {
+  val yamlPrinter = Printer.spaces2.copy(dropNullKeys = true, preserveOrder = true)
+
+  def assertCorrectJson(cwl: Cwl, expectedYaml: String): Unit = {
+    val cwlJson = CwlCodecs.encodeCwl(cwl)
+    val cwlYaml = yamlPrinter.pretty(cwlJson)
```
danbills: if you add this call as a function in  `CwlCodecs` tthen everyone will have X -> Yaml and they will be happy about that.

i.e. `encodeWorkflow(workflow) : Yaml`  might be preferable to returning Json.  It could be provided in addition to the Json function

cwl/src/main/scala/wdl4s/cwl/Cwl.scala
danbills: I see now that was formerly the case and am now curious why this change was made?

curoli: I got sporadic compiler errors telling me it found String but required String("CommandLineTool") or String("Workflow") when I tried to instantiate CommandLineTool or Workflow.

I did not always get this error when instantiating CommandLineTool or Workflow, but when I got it, I could not figure out, why or how to prevent it, except for the approach you see here. I'm suspecting the Refined library uses a macro with some odd limitations.

Either way, my reasoning is that if you want to use properly the type of a singleton, you have to create a reference to that singleton so that you can ask for its type. So this is what I did.

Thoughts?

curoli: Just tried to inline in CommandLineTool and Workflow, and I get:
```
[error] /home/oruebenacker/git/wdl4s/cwl/src/test/scala/wdl4s/cwl/ExportCwlSamplesSpec.scala:22: type mismatch;
[error]  found   : String
[error]  required: String("CommandLineTool")
[error]       CommandLineTool(
[error]       ^
[error] /home/oruebenacker/git/wdl4s/cwl/src/test/scala/wdl4s/cwl/ExportCwlSamplesSpec.scala:70: type mismatch;
[error]  found   : String
[error]  required: String("Workflow")
[error]     val workflow = Workflow(
[error]                    ^
[error] /home/oruebenacker/git/wdl4s/cwl/src/test/scala/wdl4s/cwl/ExportCwlSamplesSpec.scala:141: type mismatch;
[error]  found   : String
[error]  required: String("CommandLineTool")
[error]     val tool = CommandLineTool(
[error]                ^
[error] three errors found
[error] (cwl/test:compileIncremental) Compilation failed
[error] Total time: 190 s, completed 26.07.2017 17:10:38
```

danbills:
Huh.  I guess it could be a bug of some sort as I recall seeing something
similar.  We can go with your approach.

curoli:
Addressed all reviewers' feedback.

cwl/src/main/scala/wdl4s/cwl/Cwl.scala
danbills: Hm I tried to reply on my phone but apparently it didn't submit.  I think your workaround is fine given that I saw some weirdness too.

cwl/src/main/scala/wdl4s/cwl/Cwl.scala
```diff
@@ -13,12 +13,16 @@ sealed trait Cwl {
 }
 
 case class Workflow(
-  cwlVersion: Option[CwlVersion] = None,
-  `class`: Witness.`"Workflow"`.T = "Workflow".narrow,
+  cwlVersion: Option[CwlVersion] = Option(CwlVersion.default),
```
danbills: Sorry, I should have been explicit that the default should be set here, not in the enumeration.

Thus this would be `cwlVersion: Option[CwlVersion] = Option(CwlVersion.Version1)` and please remove "default" from the enumeration.

cwl/src/main/scala/wdl4s/cwl/Cwl.scala
```diff
@@ -42,13 +46,13 @@ case class Workflow(
...
                             outputs: Outputs = Coproduct[Outputs](Array.empty[CommandOutputParameter]),
-                            `class`: W.`"CommandLineTool"`.T = "CommandLineTool".narrow,
+                            `class`: CommandLineTool.`class`.type = CommandLineTool.`class`,
                             id: Option[String] = None,
                             requirements: Option[Array[Requirement]] = None,
                             hints: Option[Array[String]] = None, //TODO: Any?
                             label: Option[String] = None,
                             doc: Option[String] = None,
-                            cwlVersion: Option[CwlVersion] = None,
+                            cwlVersion: Option[CwlVersion] = Option(CwlVersion.default),
```
danbills: as noted above.  Please and thanks!

cwl/src/main/scala/wdl4s/cwl/Cwl.scala
curoli: I can do that, but wouldn't that be a pain if we ever switch to another version?

curoli: Ok, I changed it to CwlVersion.Version1.

(If CommandLineTool and Workflow are the only objects using it, I guess it's no big deal).

curoli:
Weird: GitHub tells me that @mcovarr and @danbills approved, but at the same time it is saying that I have zero of two required pull approves (from a list that includes @mcovarr and @danbills).

But I'm assuming I'm ready to merge.

danbills:
👍

mcovarr:
:+1:

PR #161, opened 2017-08-03
WOMify some interesting CWLs and assert things look reasonable.

mcovarr:


wom/src/main/scala/wdl4s/wom/graph/CallNode.scala
```diff
@@ -49,7 +49,9 @@ object TaskCall {
 
 object CallNode {
 
-  final case class CallWithInputs(call: CallNode, inputs: Set[GraphInputNode])
+  final case class CallWithInputs(call: CallNode, inputs: Set[GraphInputNode]) {
+    def nodes: Set[GraphNode] = Set.empty[GraphNode] + call ++ inputs
```
kshakir: ToL: Why not `Set(call) ++ inputs`?

mcovarr: I thought there was an issue with the types not working out right, but apparently not.  Will fix.

cwl/src/test/scala/wdl4s/cwl/CwlWorkflowWomSpec.scala
```diff
@@ -0,0 +1,301 @@
...
+  id: message
+outputs: []
+""".stripMargin
+
+    CwlCodecs.decodeCwl(firstTool) match {
+      case Right((clt: CommandLineTool, _)) =>
+        clt.womExecutable match {
+          case Valid(wom) =>
+            wom.entryPoint match {
+              case workflow: TaskDefinition =>
```
cjllanwarne: Is this three-`match` logic going to be needed every time anyone runs `decodeCwl`? 

My friend, Naïve Chris would say "Surely it's either valid or it isn't?"

cwl/src/main/scala/wdl4s/cwl/package.scala
```diff
@@ -53,4 +69,6 @@ package object cwl extends TypeAliases {
 
   type Yaml = String
```
cjllanwarne: I'm personally not a big fan of pretending that `String`s are `Yaml`s. IMO it just gives the false security of type safety while actually just making it harder to work out what arguments are wanted.

cwl/src/test/scala/wdl4s/cwl/CwlWorkflowWomSpec.scala
cjllanwarne: As a side note, `workflow: TaskDefinition` seems like a copy/paste has gone wrong somewhere

cwl/src/main/scala/wdl4s/cwl/package.scala
geoffjentry: You all know my take on this. Unless it's a type alias to make cats happy (which I don't think either this one or the new one below are) I'd prefer to see it as a wrapped case class even if that makes @mcovarr grumpy having to type a few extra chars before his autocomplete works.

cwl/src/test/scala/wdl4s/cwl/CwlWorkflowWomSpec.scala
```diff
@@ -0,0 +1,301 @@
...
+    }
+
+    def validateWom(ex: Executable) = {
+      ex match {
+        case Executable(wf: WorkflowDefinition) =>
+          val nodes = wf.innerGraph.nodes
+
+          nodes collect {
+            case gin: GraphInputNode => gin.name
+          } should be(Set("file:///home/dan/wdl4s/r.cwl#ex", "file:///home/dan/wdl4s/r.cwl#inp"))
```
cjllanwarne: While the file names are important for the CWL parser doing the linkings, by the time we get to WOM, it'd be nice if these are just the local/unqualified names

cjllanwarne: Otherwise, our workflow graphs are going to look like this:

outfile

cwl/src/main/scala/wdl4s/cwl/package.scala
danbills: I'll just use `String` as this is upstream of validation

cwl/src/test/scala/wdl4s/cwl/CwlWorkflowWomSpec.scala
danbills: Initially we wanted the "nice" names.  I believe the names are being used to create connections, and we couldn't ensure global uniqueness without the fully qualified name.

cwl/src/main/scala/wdl4s/cwl/package.scala
danbills: To be clear, a case class wrapping a string that has a validating constructor would be superior, but that happens as part of yaml parsing so I'm not sure about doing it now.

cwl/src/main/scala/wdl4s/cwl/WorkflowOutputsToOutputDefinition.scala
```diff
@@ -0,0 +1,31 @@
+package wdl4s.cwl
+
+import shapeless.Poly1
+import wdl4s.wom.callable.Callable.OutputDefinition
+import wdl4s.wom.expression.PlaceholderWomExpression
+
+object WorkflowOutputsToOutputDefinition extends Poly1 {
+
+  def mungeId(fullyQualifiedName: String): String =
+    fullyQualifiedName.substring(fullyQualifiedName.lastIndexOf("/") + 1)
```
cjllanwarne: `+ 1`?

cwl/src/test/scala/wdl4s/cwl/CwlWorkflowWomSpec.scala
```diff
@@ -0,0 +1,299 @@
...
+            upstream shouldBe Set(
+            RequiredGraphInputNode("file:///home/dan/wdl4s/r.cwl#ex", WdlStringType),
+            RequiredGraphInputNode("file:///home/dan/wdl4s/r.cwl#inp", WdlFileType))
+
+
+          nodes.collectFirst {
+            case compile: CallNode if compile.name == "file:///home/dan/wdl4s/r.cwl#compile" => compile
+          }.get.inputPorts.map(_.upstream).head.name shouldBe "file:///home/dan/wdl4s/r.cwl#untar/example_out"
+
+        //TODO: test for outputs
```
cjllanwarne: Is this TODONE? Or a ticket?

cwl/src/test/scala/wdl4s/cwl/CwlWorkflowWomSpec.scala
```diff
@@ -0,0 +1,299 @@
...
+        "file:///Users/danb/wdl4s/r.cwl#cgrep",
+        "file:///Users/danb/wdl4s/r.cwl#wc"))
+
+    val ps = nodes.collectFirst({ case ps: CallNode if ps.name == "file:///Users/danb/wdl4s/r.cwl#ps" => ps }).get
+    val cgrep = nodes.collectFirst({ case cgrep: CallNode if cgrep.name == "file:///Users/danb/wdl4s/r.cwl#cgrep" => cgrep }).get
+    val cgrepPatternInput = nodes.collectFirst({ case cgrepInput: GraphInputNode if cgrepInput.name == "file:///Users/danb/wdl4s/r.cwl#pattern" => cgrepInput }).get
+    val wc = nodes.collectFirst({ case wc: CallNode if wc.name == "file:///Users/danb/wdl4s/r.cwl#wc" => wc }).get
+
+    ps.upstream shouldBe empty
+    cgrep.upstream shouldBe Set(ps, cgrepPatternInput)
```
cjllanwarne: Are these `shouldBe`s based on object equality or `case class` equality?

I presume it's `case class` equality... is there any way of instead checking object equality?

wom/src/main/scala/wdl4s/wom/graph/Graph.scala
```diff
@@ -8,6 +8,8 @@ import wdl4s.wom.graph.GraphNodePort.InputPort
   * A sealed set of graph nodes.
   */
 final case class Graph private(nodes: Set[GraphNode]) {
+  // FIXME: this actually seems pretty WDL specific, and even then specific to the case of a WDL that doesn't have
+  // an explicit outputs block.  I'd remove this but the WDL/WOM test relies on it.
```
danbills: It is used both in "regular" and test code, so it can't be moved to only test code.

cwl/src/test/scala/wdl4s/cwl/ParseBigThreeSpec.scala
```diff
@@ -9,6 +9,8 @@ class ParseBigThreeSpec extends FlatSpec with Matchers {
   val namespace = "cwl"
 
   it should "parse 1st tool" in {
+    //This id:message is just boilerplate because our object requires an id
+    // and we don't support the map[] declaration yet
```
cjllanwarne: Yet? I thought that was a SALAD responsibility?

wom/src/main/scala/wdl4s/wom/graph/Graph.scala
cjllanwarne: I think it should probably move somewhere in the `wdl4s.wdl` package (or maybe WDL already has a function for this and we should use that instead).

FWIW I think it's more likely to get fixed with a ticket than a code FIXME, but that's not a strong opinion either way

cwl/src/test/scala/wdl4s/cwl/ParseBigThreeSpec.scala
danbills: Map declarations are indeed supported by SALAD.  deleting this comment as it no longer applies

cwl/src/main/scala/wdl4s/cwl/WorkflowOutputsToOutputDefinition.scala
danbills: this code is orphaned and will be deleted

cwl/src/test/scala/wdl4s/cwl/CwlWorkflowWomSpec.scala
mcovarr: `case class`es synthesize an `equals` method which check value equality.  `eq` will check reference equality.

PR #164, opened 2017-08-09
WomExpression structure and ExpressionBasedGraphOutputNode

cjllanwarne:
The graph generated in `GraphOutputNodeSpec` is graphically this:
test


geoffjentry:
@cjllanwarne Which ticket(s) does this close, if any?

cjllanwarne:
@geoffjentry #2522

It helps unblock @mcovarr to look at `WDL => WomExpression` and opens up my next course of action as adding DeclarationNode definitions in WOM

mcovarr:
Could you verbally describe what's going on in that picture?

cjllanwarne:
@mcovarr sure (although as a side note, the fact that you asked makes me wonder if I can diagram expressions better...)

- `i` and `j` are `GraphInputNode`s, representing graph inputs. 
- `out_j` is a `PortBasedGraphOutputNode`, ie an output that is just exposing an existing `OutputPort` as a graph output.
- `out_x` is an `ExpressionBasedGraphOutputNode`, which is a graph output built on top of an expression that (in this case), takes `i` and `j` inputs and creates an `Int x` output.

In WDL this might look something like this (although the expression could be anything):
```
workflow foo {
  Int i
  Int j
  output {
    Int j_out = j
    Int x_out = i + j
  }
}
```

mcovarr:
Actually the more I'm looking at this I wonder if it wouldn't be simpler to make output port based graph output nodes just a degenerate form of expression based graph output nodes rather than having two different types.

cjllanwarne:
@mcovarr I suspect that's definitely fine in WDL (and probably how wdl4s would build WOM).

I'm not sure how CWL will end up using these however, since I think CWL natively talks the language of exposing ports (and running JS just to mirror an input might be OTT)?

wom/src/main/scala/wdl4s/wom/expression/WomExpression.scala
```diff
@@ -0,0 +1,31 @@
...
+final case class NamedExpressionInput(name: String, womType: WdlType)
+
+final case class ExpressionInputs(values: Map[NamedExpressionInput, WdlValue]) extends AnyVal
+
+trait IoFunctionSet {
+  def read_file(path: String): Future[String]
+  def write_file(path: String, content: String): Future[Unit]
+}
+
+case object PlaceholderIoFunctionSet {
```
danbills: is this perhaps supposed to `extend IoFunctionSet`?

wom/src/main/scala/wdl4s/wom/graph/GraphOutputNode.scala
```diff
@@ -1,9 +1,53 @@
...
+final case class ExpressionBasedGraphOutputNode private(name: String, expression: WomExpression, inputMapping: Map[NamedExpressionInput, InputPort]) extends GraphOutputNode {
+  override val womType = expression.womType
+  override val inputPorts = inputMapping.values.toSet
+}
+
+object ExpressionBasedGraphOutputNode {
+  def linkWithInputs(name: String, expression: WomExpression, inputMapping: Map[String, OutputPort]): ErrorOr[ExpressionBasedGraphOutputNode] = {
+    val graphNodeSetter = new GraphNode.GraphNodeSetter()
+
+    def linkInput(input: NamedExpressionInput): ErrorOr[(NamedExpressionInput, InputPort)] = if (inputMapping.contains(input.name)) {
```
danbills: ToL put `if` on new line?

PR #165, opened 2017-08-10
WOM ExpressionNodes

cjllanwarne:
- [x] To be rebased on develop after https://github.com/broadinstitute/wdl4s/pull/164 merges.

Easiest entry point for reviewers: the `ExpressionNodeSpec` creates the graph:

test

i.e. roughly equivalent to:
```
Int i
Int j
Int x = i + j # Or could be any expression of i and j
output {
  Int x_out = x
}
```

### DIFF
--- a/wom/src/main/scala/wdl4s/wom/graph/InstantiatedExpression.scala
+++ b/wom/src/main/scala/wdl4s/wom/graph/InstantiatedExpression.scala
@@ -1,0 +1,44 @@
+package wdl4s.wom.graph
+
+import cats.instances.list._
+import cats.syntax.traverse._
+import cats.syntax.validated._
+import cats.data.Validated.Valid
+import lenthall.validation.ErrorOr.ErrorOr
+import wdl4s.wdl.types.WdlType
+import wdl4s.wom.expression.WomExpression
+import wdl4s.wom.graph.GraphNodePort.{ConnectedInputPort, InputPort, OutputPort}
+
+class InstantiatedExpression private(val expression: WomExpression, val womReturnType: WdlType, val inputMapping: Map[String, InputPort]) {
+  val inputPorts = inputMapping.values.toSet
+}
+
+object InstantiatedExpression {
+
+  private[graph] def instantiateExpressionForNode[ExpressionBasedNode <: GraphNode](nodeConstructor: (String, InstantiatedExpression) => ExpressionBasedNode)(name: String, expression: WomExpression, inputMapping: Map[String, OutputPort]): ErrorOr[ExpressionBasedNode] = {
+    val graphNodeSetter = new GraphNode.GraphNodeSetter()
+
+    for {
+      linkedInputs <- InstantiatedExpression.linkWithInputs(graphNodeSetter, expression, inputMapping)
+      expressionNode = nodeConstructor(name, linkedInputs)
+      _ = graphNodeSetter._graphNode = expressionNode
+    } yield expressionNode
+  }
+
+  def linkWithInputs(graphNodeSetter: GraphNode.GraphNodeSetter, expression: WomExpression, inputMapping: Map[String, OutputPort]): ErrorOr[InstantiatedExpression] = {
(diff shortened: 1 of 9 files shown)


Horneth:
ToL: Is there going to be a declaration node ? If yes, what will be the difference with ExpressionNode ?

cjllanwarne:
@Horneth this probably will be what we use for WDL expressions.

For other expressions (eg expressions in call inputs, expressions for scatter variables), I'm in two minds whether to have them:
- As separate, anonymous, expression nodes
- Built into the nodes that might use them.

I *think* I'm leaning more towards the latter right now but otoh I think it might be easier to have them separate.

Horneth:
@cjllanwarne So a WDL task output expression would translate to a WomExpressionNode ?
For the other ones I'm not sure either. Having a node *wrapped* in another node seems a bit odd but maybe it's not.

cjllanwarne:
@Horneth probably easier to explain in person but I don't think that should be necessary!

Re the wrapping, we're going to have to create one of these `InstantiatedExpression` things for call inputs, the question is whether they're stored in the same GraphNode that owns the expression or as a separate ExpressionNode that feeds into it

Horneth:
👍

mcovarr:
I'm thinking separate anonymous nodes could create yuckiness for {wom|wdl}tool and anyone inspecting a graph by other means.  Knowing what I know now (which admittedly isn't all that much) I'd lean toward encapsulation in the owning node.

wom/src/main/scala/wdl4s/wom/graph/InstantiatedExpression.scala
```diff
@@ -0,0 +1,44 @@
...
+import wdl4s.wom.expression.WomExpression
+import wdl4s.wom.graph.GraphNodePort.{ConnectedInputPort, InputPort, OutputPort}
+
+class InstantiatedExpression private(val expression: WomExpression, val womReturnType: WdlType, val inputMapping: Map[String, InputPort]) {
+  val inputPorts = inputMapping.values.toSet
+}
+
+object InstantiatedExpression {
+
+  private[graph] def createExpressionNodeLinkedWithInputs[A <: GraphNode](applier: (String, InstantiatedExpression) => A)(name: String, expression: WomExpression, inputMapping: Map[String, OutputPort]): ErrorOr[A] = {
```
mcovarr: Kind of surprising `A` can't be constrained any more than as a subtype of `GraphNode`.  It seems like there's a missing type there, even if it's just a marker trait.

wom/src/main/scala/wdl4s/wom/graph/InstantiatedExpression.scala
```diff
@@ -0,0 +1,44 @@
...
+
+    import lenthall.validation.ErrorOr.ShortCircuitingFlatMap
+    for {
+      linkedInputList <- expression.inputs.toList traverse linkInput
+      linkedInputs = linkedInputList.toMap
+      inputTypes = linkedInputs map { case (k, v) => k -> v.womType }
+      evaluatedType <- expression.womType(inputTypes)
+    } yield new InstantiatedExpression(expression, evaluatedType, linkedInputs)
+  }
+}
```
mcovarr: 🚫

wom/src/test/scala/wdl4s/wom/graph/ExpressionNodeSpec.scala
```diff
@@ -0,0 +1,62 @@
...
+    graph match {
+      case Valid(g) => validate(g)
+      case Invalid(errors) => fail(s"Unable to build WOM graph: ${errors.toList.mkString(" ")}")
+    }
+
+    def validate(graph: Graph) = {
+      val x_outGraphOutputNode = graph.nodes.find {
+        case g: GraphOutputNode => g.name == "x_out"
+        case _ => false
+      }.getOrElse(fail("Verification couldn't find the x_out GraphOutputNode"))
```
mcovarr: who's "Verification"?

PR #169, opened 2017-08-11
Coerce JsNull to WdlOptionalValue.none

Horneth:


geoffjentry:
Should this find its way into the spec or is it purely an implementation detail?

And if the former, I bring up my typical request that any change to the WDL spec (I mean in spirit, not actual doc) be put in front of the wdl focus group, in particular @vdauwera

Horneth:
@geoffjentry I would argue that this is what it should have been doing in the first place and could even be considered a bug but it's just my opinion. If you think it should be wdl-group-focused I'll post it in the room :)

ruchim:
:+1: :1st_place_medal:

PR #172, opened 2017-08-15
publish release script using version numbers w/o git hash

danbills:


danbills:
For the record I tested this locally and on travis server.

PR #180, opened 2017-08-19
WdlWomExpressions

mcovarr:
Many thanks to @cjllanwarne for his WDL-wizardry.  Lessons were learned:

* Several required functions were missing from `IoFunctionSet`.
* The signatures of `IoFunctionSet` functions were wisely designed to return `Future[T]`.  The signatures of WDL functions return `T`.  Guess what the WDL implementations do to bridge this difference.  :nauseated_face: 
* Warning: changes in `TypeEvaluator` are more disgusting than they appear.  Basically we try a bunch of convoluted lookup logic and if that fails, eh just consult the lookup function that was passed in instead.  That will probably break down for expressions in scatters or conditionals.
* The cleanliness of the `IoFunctionSet` design inspired / necessitated some Cromwell refactoring, represented in a forthcoming Cromwell PR.

### DIFF
--- a/wom/src/main/scala/wdl4s/wdl/AstTools.scala
+++ b/wom/src/main/scala/wdl4s/wdl/AstTools.scala
@@ -305,6 +305,8 @@ object AstTools {
         && a.getAttribute("lhs") == terminal
         && a.getAttribute("rhs").isTerminal => a.getAttribute("rhs").asInstanceOf[Terminal]
     }
+
+    def fullVariableReferenceString = terminal.getSourceString + (terminalSubIdentifier map { "." + _.getSourceString } getOrElse "")
   }
 
   /**
--- a/wom/src/main/scala/wdl4s/wdl/WdlCall.scala
+++ b/wom/src/main/scala/wdl4s/wdl/WdlCall.scala
@@ -1,11 +1,12 @@
 package wdl4s.wdl
 
 import cats.data.Validated.Valid
-import cats.syntax.traverse._
 import cats.instances.list._
 import cats.syntax.cartesian._
-
-import lenthall.validation.ErrorOr.ErrorOr
+import cats.syntax.option._
+import cats.syntax.traverse._
+import cats.syntax.validated._
+import lenthall.validation.ErrorOr.{ErrorOr, ShortCircuitingFlatMap}
 import wdl4s.parser.WdlParser.{Ast, SyntaxError, Terminal}
 import wdl4s.wdl.AstTools.EnhancedAstNode
 import wdl4s.wdl.exception.{ValidationException, VariableLookupException, VariableNotFoundException}
@@ -13,9 +14,7 @@ import wdl4s.wdl.expression.WdlFunctions
 import wdl4s.wdl.types.WdlOptionalType
 import wdl4s.wdl.values.{WdlOptionalValue, WdlValue}
 import wdl4s.wom.graph.CallNode.CallWithInputs
-import wdl4s.wom.graph.{CallNode, GraphInputNode, GraphNodePort}
-import cats.syntax.validated._
-import lenthall.validation.ErrorOr.ShortCircuitingFlatMap
+import wdl4s.wom.graph._
 
 import scala.language.postfixOps
 import scala.util.{Failure, Success, Try}
@@ -56,28 +55,35 @@ object WdlCall {
 
   private def buildWomNodeAndInputs(wdlCall: WdlCall): ErrorOr[CallWithInputs] = {
 
-    val inputToOutputPortValidations: Iterable[ErrorOr[(String, GraphNodePort.OutputPort)]] = for {
-      (inputName, expr) <- wdlCall.inputMappings
-      variable <- expr.variableReferences
-      parent <- wdlCall.parent
-      node <- parent.resolveVariable(variable.terminal.sourceString)
-      outputPort = outputPortFromNode(node, variable.terminalSubIdentifier)
-      inputNameToOutputPortMapping = outputPort map { op => inputName -> op }
-    } yield inputNameToOutputPortMapping
+    def resolveVariable(v: AstTools.VariableReference): ErrorOr[(String, GraphNodePort.OutputPort)] = {
+      val name =  v.fullVariableReferenceString
+      for {
+        parent <- wdlCall.parent.toValidNel("Call has no parent?")
+        node <- parent.resolveVariable(v.terminal.sourceString).toValidNel(s"Failed to resolve variable $name")
+        outputPort <- outputPortFromNode(node, v.terminalSubIdentifier)
+      } yield name -> outputPort
+    }
 
-    val inputToOutputPortValidation: ErrorOr[List[(String, GraphNodePort.OutputPort)]] = inputToOutputPortValidations.toList.sequence
 
-    // TODO: When WomExpressions wrap WdlExpressions, we can switch this to use expressionBasedInputs instead of portBasedInputs:
-    val combinedValidation = (inputToOutputPortValidation |@| wdlCall.callable.womDefinition) map { (_, _) }
+    val graphNodeInputExpressionValidation: ErrorOr[List[GraphNodeInputExpression]] = {
+      val graphNodeInputExpression: ((String, WdlExpression)) => ErrorOr[GraphNodeInputExpression] = {
+        case (name, wdlExpression) =>
+          for {
+            resolvedVariables <- wdlExpression.variableReferences.toList traverse resolveVariable
+            womExpression = WdlWomExpression(wdlExpression, None)
+          } yield GraphNodeInputExpression(name, womExpression, resolvedVariables.toMap)
+      }
+
+      wdlCall.inputMappings.toList traverse graphNodeInputExpression
+    }
 
     for {
-      combined <- combinedValidation
+      combined <- (graphNodeInputExpressionValidation |@| wdlCall.callable.womDefinition).tupled
       (inputToOutputPort, callable) = combined
-      callWithInputs <- CallNode.callWithInputs(wdlCall.alias.getOrElse(wdlCall.callable.unqualifiedName), callable, inputToOutputPort.toMap, Set.empty)
+      callWithInputs <- CallNode.callWithInputs(wdlCall.alias.getOrElse(wdlCall.callable.unqualifiedName), callable, Map.empty, inputToOutputPort.toSet)
     } yield callWithInputs
   }
 
-  //
   private def outputPortFromNode(node: WdlGraphNode, terminal: Option[Terminal]): ErrorOr[GraphNodePort.OutputPort] = {
 
     def findNamedOutputPort(name: String, graphOutputPorts: Set[GraphNodePort.OutputPort], terminalName: String): ErrorOr[GraphNodePort.OutputPort] = {
--- a/wom/src/main/scala/wdl4s/wdl/WdlExpression.scala
+++ b/wom/src/main/scala/wdl4s/wdl/WdlExpression.scala
@@ -1,5 +1,7 @@
 package wdl4s.wdl
 
+import lenthall.validation.ErrorOr.ErrorOr
+import lenthall.validation.Validation._
 import wdl4s.parser.WdlParser
 import wdl4s.parser.WdlParser.{Ast, AstList, AstNode, Terminal}
 import wdl4s.wdl.AstTools.{EnhancedAstNode, VariableReference}
@@ -8,8 +10,11 @@ import wdl4s.wdl.expression._
 import wdl4s.wdl.formatter.{NullSyntaxHighlighter, SyntaxHighlighter}
 import wdl4s.wdl.types._
 import wdl4s.wdl.values._
+import wdl4s.wom.expression.{IoFunctionSet, WomExpression}
 
 import scala.collection.JavaConverters._
+import scala.concurrent.Await
+import scala.concurrent.duration.Duration
 import scala.language.postfixOps
 import scala.util.Try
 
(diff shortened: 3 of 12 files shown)


wom/src/main/scala/wdl4s/wdl/WdlExpression.scala
```diff
@@ -186,6 +192,43 @@ case class WdlExpression(ast: AstNode) extends WdlValue {
...
+  */
+final case class WdlWomExpression(wdlExpression: WdlExpression, from: Option[Scope]) extends WomExpression {
+
+  override def inputs: Set[String] = wdlExpression.variableReferences map { _.fullVariableReferenceString } toSet
+
+  override def evaluateValue(variableValues: Map[String, WdlValue], ioFunctionSet: IoFunctionSet): ErrorOr[WdlValue] = {
+    lazy val wdlFunctions = new WdlStandardLibraryFunctions {
+      override def readFile(path: String): String = Await.result(ioFunctionSet.readFile(path), Duration.Inf)
+
+      override def writeFile(path: String, content: String): Try[WdlFile] = Try(Await.result(ioFunctionSet.writeFile(path, content), Duration.Inf))
```
Horneth, line 208: Is there a reason why some of those functions return a `Try[...]` but not all ?

wom/src/main/scala/wdl4s/wdl/WdlExpression.scala
```diff
@@ -186,6 +192,43 @@ case class WdlExpression(ast: AstNode) extends WdlValue {
...
+
+      override def stdout(params: Seq[Try[WdlValue]]): Try[WdlFile] = ???
```
Horneth: ToL (and not for this PR anyway): I'm wondering if it wouldn't make more sense to pass stderr and stdout as part of the "value context" or whatever name we give the other thing we pass to the `evaluate` method mapping variables to values. It happens that in wdl stdout and stderr are functions but really they're static values not requiring any I/O nor computation I think.

wom/src/main/scala/wdl4s/wdl/expression/TypeEvaluator.scala
```diff
@@ -96,6 +96,8 @@ case class TypeEvaluator(override val lookup: String => WdlType, override val fu
               }
             case ns: WdlNamespace => Success(lookup(ns.importedAs.map{ n => s"$n.${rhs.getSourceString}" }.getOrElse(rhs.getSourceString)))
             case _ => Failure(new WdlExpressionException("Left-hand side of expression must be a WdlObject or Namespace"))
+          } recoverWith {
+            case _ => Try(lookup(a.getAttribute("lhs").sourceString + "." + rhs.sourceString))
```
Horneth, line 100: Just curious, what requires this that the logic above can't evaluate ?

Horneth (approved):

wom/src/main/scala/wdl4s/wdl/WdlExpression.scala
Horneth: Oh those are the WDL functions not wom, nevermind, I need more coffee -_-

wom/src/main/scala/wdl4s/wdl/expression/TypeEvaluator.scala
cjllanwarne: Up until now, when the `TypeEvaluator` is created it was given a lookup function that works for everything except member accesses (which it gets in some convoluted way from the `from` parameter).

For WOM, we don't need the `from` and can just pass in a fully-stocked lookup (since we already know which lookups the evaluators are going to do). So we let the `from`-based lookup fail and fall back to the `lookup`-based lookup here.

If this survives the coming `scatter` shakeup then I think we'll probably want to parameterize how to do member access lookups rather than trying both and falling back if the first way fails

wom/src/main/scala/wdl4s/wdl/WdlCall.scala
```diff
@@ -56,28 +54,41 @@ object WdlCall {
 
   private def buildWomNodeAndInputs(wdlCall: WdlCall): ErrorOr[CallWithInputs] = {
 
-    val inputToOutputPortValidations: Iterable[ErrorOr[(String, GraphNodePort.OutputPort)]] = for {
+    def resolvedVariableToErrorOr(name: String, v: Option[WdlGraphNode]): ErrorOr[WdlGraphNode] = v match {
+      case Some(wgn) => Valid(wgn)
+      case None => s"Failed to resolve variable $name".invalidNel
+    }
```
danbills: ToL cats `Option.toInvalidNel` lets you just specify the error case

wom/src/main/scala/wdl4s/wdl/AstTools.scala
```diff
@@ -305,6 +305,8 @@ object AstTools {
         && a.getAttribute("lhs") == terminal
         && a.getAttribute("rhs").isTerminal => a.getAttribute("rhs").asInstanceOf[Terminal]
     }
+
+    def fullVariableReferenceString = terminal.getSourceString + terminalSubIdentifier.map("." + _.getSourceString).getOrElse("")
```
danbills: ToL cats' `Option.fold` is nice for the `map.getOrElse` scenario

wom/src/main/scala/wdl4s/wdl/WdlCall.scala
```diff
@@ -56,28 +54,41 @@ object WdlCall {
...
+
+    def resolveVariable(v: AstTools.VariableReference): ErrorOr[(String, GraphNodePort.OutputPort)] = {
+      val parent = wdlCall.parent.get
```
danbills: this should be part of the forcomp via `toInvalidNel`

wom/src/main/scala/wdl4s/wdl/WdlExpression.scala
```diff
@@ -186,6 +192,43 @@ case class WdlExpression(ast: AstNode) extends WdlValue {
...
+    }
+    wdlExpression.evaluate(variableValues.apply, wdlFunctions).toErrorOr
+  }
+
+  override def evaluateType(inputTypes: Map[String, WdlType]): ErrorOr[WdlType] = {
+    // All current usages of WdlExpression#evaluateType trace back to WdlNamespace, but this is not necessarily the
+    // case in the brave new WOM-world.
+    wdlExpression.evaluateType(inputTypes.apply, new WdlStandardLibraryFunctionsType, from) match {
+      case Success(t) => Valid(t)
+      case Failure(e) => throw e
```
danbills: why not Invalid?

wom/src/main/scala/wdl4s/wdl/expression/WdlStandardLibraryFunctions.scala
```diff
@@ -11,26 +11,50 @@ import wdl4s.wdl.{TsvSerializable, WdlExpressionException}
...
+
   def glob(path: String, pattern: String): Seq[String]
-  def write_tsv(params: Seq[Try[WdlValue]]): Try[WdlFile]
-  def write_json(params: Seq[Try[WdlValue]]): Try[WdlFile]
+
   def size(params: Seq[Try[WdlValue]]): Try[WdlFloat]
 
+  private def writeContent(baseName: String, content: String): Try[WdlFile] = {
+    writeFile(s"${baseName}_${content.md5Sum}.tmp", content)
+  }
```
danbills: toL no need for braces

wom/src/main/scala/wdl4s/wdl/WdlCall.scala
```diff
@@ -56,28 +54,41 @@ object WdlCall {
...
-      node <- parent.resolveVariable(variable.terminal.sourceString)
-      outputPort = outputPortFromNode(node, variable.terminalSubIdentifier)
-      inputNameToOutputPortMapping = outputPort map { op => inputName -> op }
-    } yield inputNameToOutputPortMapping
+      womExpression = WdlWomExpression(expr, None)
+      uninstantiatedExpression = graphNodeInputExpression(inputName, womExpression)
+    } yield uninstantiatedExpression
 
-    val inputToOutputPortValidation: ErrorOr[List[(String, GraphNodePort.OutputPort)]] = inputToOutputPortValidations.toList.sequence
+    val graphNodeInputExpressionValidation: ErrorOr[List[GraphNodeInputExpression]] = graphNodeInputExpressionValidations.toList.sequence
```
danbills: traverse on wdlCall.inputMappings could let you skip this step

wom/src/main/scala/wdl4s/wdl/WdlExpression.scala
```diff
@@ -186,6 +192,43 @@ case class WdlExpression(ast: AstNode) extends WdlValue {
...
+
+      override def writeFile(path: String, content: String): Try[WdlFile] = Try(Await.result(ioFunctionSet.writeFile(path, content), Duration.Inf))
+
+      override def stdout(params: Seq[Try[WdlValue]]): Try[WdlFile] = ???
+
+      override def stderr(params: Seq[Try[WdlValue]]): Try[WdlFile] = ???
+
+      override def glob(path: String, pattern: String): Seq[String] = ???
+
+      override def size(params: Seq[Try[WdlValue]]): Try[WdlFloat] = ???
```
danbills: I'm not familiar enough with what's going on here but are these todo's or are they purposefully unimplemented?  If so perhaps a comment to help the uninitiated?

wom/src/main/scala/wdl4s/wdl/WdlCall.scala
mcovarr: nice, though this one is actually `toValidNel`

wom/src/main/scala/wdl4s/wdl/WdlExpression.scala
mcovarr: oops these were stubs before I added the methods to `IoFunctionSet`, I'll update to have them call those methods.

wom/src/main/scala/wdl4s/wdl/WdlExpression.scala
mcovarr: I believe this is technically what one would describe as an oopsie.

wom/src/main/scala/wdl4s/wdl/expression/WdlStandardLibraryFunctions.scala
mcovarr: I'll Invisalign this.

wom/src/main/scala/wdl4s/wdl/AstTools.scala
```diff
@@ -306,7 +306,7 @@ object AstTools {
         && a.getAttribute("rhs").isTerminal => a.getAttribute("rhs").asInstanceOf[Terminal]
     }
 
-    def fullVariableReferenceString = terminal.getSourceString + terminalSubIdentifier.map("." + _.getSourceString).getOrElse("")
+    def fullVariableReferenceString = terminal.getSourceString + terminalSubIdentifier.fold(""){ "." + _.getSourceString }
```
cjllanwarne: ToL: I could understand what this was doing a lot faster the old way

geoffjentry: Likewise. I've always liked the idea of `fold` in these situations better than the reality. OTOH it's more typesafe so ¯\_(ツ)_/¯

mcovarr: I'm shrug on this but majority rules so I'll revert.

wom/src/main/scala/wdl4s/wdl/AstTools.scala
```diff
@@ -305,6 +305,8 @@ object AstTools {
         && a.getAttribute("lhs") == terminal
         && a.getAttribute("rhs").isTerminal => a.getAttribute("rhs").asInstanceOf[Terminal]
     }
+
+    def fullVariableReferenceString = terminal.getSourceString + (terminalSubIdentifier map { "." + _.getSourceString } getOrElse "")
```
danbills, line 309: toL can `fold` on `terminalSubIdentifier`

wom/src/main/scala/wdl4s/wdl/WdlCall.scala
```diff
@@ -56,28 +55,37 @@ object WdlCall {
...
+        outputPort <- outputPortFromNode(node, v.terminalSubIdentifier)
+      } yield name -> outputPort
+    }
 
-    val inputToOutputPortValidation: ErrorOr[List[(String, GraphNodePort.OutputPort)]] = inputToOutputPortValidations.toList.sequence
 
-    // TODO: When WomExpressions wrap WdlExpressions, we can switch this to use expressionBasedInputs instead of portBasedInputs:
-    val combinedValidation = (inputToOutputPortValidation |@| wdlCall.callable.womDefinition) map { (_, _) }
+    val graphNodeInputExpressionValidation: ErrorOr[List[GraphNodeInputExpression]] = {
+      def graphNodeInputExpression(tuple: (String, WdlExpression)): ErrorOr[GraphNodeInputExpression] = {
```
danbills: ToL I know you aren't a fan of returning 1-arg functions but here it could simplify tuple assignment.

```
... = {
case (name, wdlExpression) => ...
}
```

wom/src/main/scala/wdl4s/wdl/WdlCall.scala
```diff
@@ -56,28 +55,37 @@ object WdlCall {
...
+        for {
+          resolvedVariables <- wdlExpression.variableReferences.toList traverse resolveVariable
+          womExpression = WdlWomExpression(wdlExpression, None)
+        } yield GraphNodeInputExpression(name, womExpression, resolvedVariables.toMap)
+      }
+
+      wdlCall.inputMappings.toList traverse graphNodeInputExpression
+    }
+
+    val combinedValidation = (graphNodeInputExpressionValidation |@| wdlCall.callable.womDefinition).tupled
```
danbills: ToL could inline below?  pretty readable

wom/src/main/scala/wdl4s/wdl/WdlCall.scala
```diff
@@ -56,28 +55,37 @@ object WdlCall {
...
 
     for {
       combined <- combinedValidation
       (inputToOutputPort, callable) = combined
```
danbills: ToL (I know this is old code) I believe you can directly assign tuples in for comprehension, i.e. combine this line and above.

wom/src/main/scala/wdl4s/wom/expression/WomExpression.scala
```diff
@@ -14,17 +15,25 @@ trait WomExpression {
...
+  def stdout(params: Seq[Try[WdlValue]]): Try[WdlFile]
+  def stderr(params: Seq[Try[WdlValue]]): Try[WdlFile]
+  def glob(path: String, pattern: String): Seq[String]
+  def size(params: Seq[Try[WdlValue]]): Try[WdlFloat]
 }
 
 case object PlaceholderIoFunctionSet extends IoFunctionSet {
-  override def read_file(path: String): Future[String] = Future.successful("35")
-  override def write_file(path: String, content: String): Future[Unit] = Future.successful(())
+  override def readFile(path: String): Future[String] = Future.successful("35")
```
danbills: ToL why not `???` as below?

wom/src/main/scala/wdl4s/wom/graph/InstantiatedExpression.scala
```diff
@@ -30,7 +30,7 @@ object InstantiatedExpression {
       val upstreamPort = inputMapping(input)
       Valid((input, ConnectedInputPort(input, upstreamPort.womType, upstreamPort, graphNodeSetter.get)))
     } else {
-      s"Expression cannot be connected without the input $input.".invalidNel
+      s"Expression cannot be connected without the input $input (provided: ${inputMapping.toString})".invalidNel
```
danbills, line 33: 💯

danbills (approved):

wom/src/main/scala/wdl4s/wdl/AstTools.scala
mcovarr: I did that in the previous commit but @geoffjentry and @cjllanwarne asked me to knock it off so I did.

wom/src/main/scala/wdl4s/wdl/AstTools.scala
danbills: 😞  It's shorter and typesafe.  I don't think unfamiliarity is a valid argument, especially as this use case is not obscure and thus `fold` could be used regularly.

wom/src/main/scala/wdl4s/wdl/WdlCall.scala
mcovarr: I'm sure it's purely coincidence that my IntelliJ crashed when I tried to do this.

wom/src/main/scala/wdl4s/wom/expression/WomExpression.scala
mcovarr: no good reason, I probably should.

wom/src/main/scala/wdl4s/wdl/AstTools.scala
geoffjentry: I don't personally all that much, like i said I like the idea of `fold`s like this a lot on paper, I just aesthetically prefer the other. Either way is fine by me though

wom/src/main/scala/wdl4s/wdl/WdlCall.scala
cjllanwarne: I got issues when I tried this before. 

I can't remember exactly but along the lines of "\<something\> doesn't implement `filter`"?

mcovarr: yeah I'm seeing that now too

PR #183, opened 2017-08-22
Incorporate SALAD pre-processing of CWL

danbills:
* pre-processing all CWL files before parsing them.
 * Parsing CWL returns a single CWL object that has de-referenced linked CWL files into embedded Workflow or CommandLineTools

Some more minor stuff:

 * Switched `Cwl` from a sealed trait to a Coproduct.  This is more consistent with the rest of the model and happens to be easier to parse.

TODO: 

- [x] Make travis test this using the "cwltool" binary executable
- [x] Documentation!


cwl/src/main/scala/wdl4s/cwl/AddEmbeddedCwl.scala
```diff
@@ -0,0 +1,50 @@
...
+          wf.steps.toList.
+            flatMap(_.run.select[String].toList).
+            traverse[ParseValidated, (String, Cwl)] {
+              path =>
+                CwlDecoder.
+                  decodeAllCwl(Path(path.drop(5))).
+                  map(path -> _).value.map(_.toValidated)
+            }(Applicative[IO] compose Applicative[ErrorOr]).
+
+            map(_.map(_.toMap)).
```
danbills: ew

cwl/src/main/scala/wdl4s/cwl/AddEmbeddedCwl.scala
```diff
@@ -0,0 +1,50 @@
...
+            map(_.toEither)
+        }.map{
+          map =>
```
danbills: too many maps, let's get creative with this arg name

cwl/src/main/scala/wdl4s/cwl/AddEmbeddedCwl.scala
```diff
@@ -0,0 +1,50 @@
...
+import cats.effect.IO
+import CwlDecoder.{Parse, ParseValidated}
+import cats.{Applicative, Monad}
+import cats.data.EitherT._
+import cats.data.EitherT
+import lenthall.validation.ErrorOr._
+
+object AddEmbeddedCwl extends Poly1 {
+  val workflowStepLens = lens[Workflow].steps
+  val workflowStepRunLens = lens[WorkflowStep].run
```
danbills: these probably don't need to be declared at this high of a level

cwl/src/main/scala/wdl4s/cwl/Decoder.scala
```diff
@@ -0,0 +1,52 @@
...
+import ammonite.ops.ImplicitWd._
+import cats.effect.IO
+import cats.syntax.either._
+
+import scala.util.Try
+
+object CwlDecoder {
+
+  type Parse[A] = EitherT[IO, NonEmptyList[String], A]
+  type ParseValidated[A] = IO[ValidatedNel[String, A]]
```
danbills: this is used elsewhere, should probably move it there.

cwl/src/main/scala/wdl4s/cwl/EmbeddedFileNames.scala
```diff
@@ -0,0 +1,29 @@
+
```
danbills: whoops, this was meant to be deleted

cwl/src/main/scala/wdl4s/cwl/RunOutputsToTypeMap.scala
```diff
@@ -29,22 +29,25 @@ object RunOutputsToTypeMap extends Poly1 {
...
           handleCommandLine(clt)
     }
 
   implicit def string = at[String] {
     fileName =>
-      (cwlMap: Map[String, CwlFile]) =>
+      (cwlMap: Map[String, Cwl]) =>
         cwlMap(fileName) match {
+          case _ => Map.empty[String, WdlType]
+          /*
```
danbills: delete me

cwl/src/main/scala/wdl4s/cwl/WorkflowStep.scala
```diff
@@ -50,9 +52,9 @@ case class WorkflowStep(
     }.toSet
 
 
-  def taskDefinitionOutputs(cwlMap: Map[String, CwlFile]): Set[Callable.OutputDefinition] = {
+  def taskDefinitionOutputs: Set[Callable.OutputDefinition] = {
 
-    val runnableIdToTypeMap: WdlTypeMap = run.fold(RunOutputsToTypeMap).apply(cwlMap)
+    val runnableIdToTypeMap: WdlTypeMap = run.fold(RunOutputsToTypeMap).apply(Map.empty)
```
danbills: This should go away as we don't need the Cwl Map anymore

cwl/src/main/scala/wdl4s/cwl/WorkflowStep.scala
```diff
@@ -179,6 +179,24 @@ case class WorkflowStepOutput(id: String)
 
 object WorkflowStep {
 
+  val emptyOutputs: Outputs = Coproduct[Outputs](Array.empty[String])
+
+  def validate(
```
danbills: this isn't a good name.  Just a constructor

cwl/src/test/resources/bad2.cwl
```diff
@@ -0,0 +1 @@
+gibberish
```
cjllanwarne, line 1: How is `bad2.cwl` from `bad.cwl`?

cjllanwarne (commented):
Curious what the scope of this PR is?

I thought you were planning to do the CWL salad as a `ServiceRegistry` service (which I may have missed here?) Or are we now planning to handle all CWL (salad-ed or not) with scala methods?

cwl/src/main/scala/wdl4s/cwl/AddEmbeddedCwl.scala
mcovarr: map

cwl/src/main/scala/wdl4s/cwl/AddEmbeddedCwl.scala
mcovarr: honestly I'm so lost I don't even know what to suggest, but yeah 😦

cwl/src/main/scala/wdl4s/cwl/CwlEncoder.scala
```diff
@@ -0,0 +1,9 @@
+package wdl4s.cwl
+
+import shapeless._
+
+object CwlEncoder extends Poly1 {
+  implicit def workflow = at[Workflow]{ CwlCodecs.encodeCwlWorkflow(_) }
```
mcovarr: not only *can* these be converted to method values, but doing so makes IntelliJ stop yellowing

project/Dependencies.scala
```diff
@@ -1,7 +1,7 @@
 import sbt._
 
 object Dependencies {
-  val catsV = "0.9.0"
+  val catsV = "1.0.0-MF"
```
mcovarr, line 4: 😸

project/Dependencies.scala
```diff
@@ -50,8 +51,13 @@ object Dependencies {
...
     "io.circe" %% "circe-yaml" % "0.6.1",
     "eu.timepit" %% "refined"            % "0.8.2",
-    "com.lihaoyi" %% "ammonite-ops" % "1.0.0-RC7" % "test",
-    "org.pegdown" % "pegdown" % "1.6.0" % Test
+    "com.lihaoyi" %% "ammonite-ops" % "1.0.1",
+    "org.typelevel" %% "cats-effect" % "0.4",
+    "org.pegdown" % "pegdown" % "1.6.0" % Test,
+     "org.scalactic" %% "scalactic" % "3.0.1",
+ "org.scalatest" %% "scalatest" % "3.0.2" % "test",
+    "org.scalacheck" %% "scalacheck" % "1.13.4" % "test"
```
mcovarr: columnarity

wom/src/main/scala/wdl4s/wdl/expression/WdlStandardLibraryFunctions.scala
```diff
@@ -174,7 +174,7 @@ trait WdlStandardLibraryFunctions extends WdlFunctions[WdlValue] {
 
   def prefix(params: Seq[Try[WdlValue]]): Try[WdlArray] = {
     def extractTwoArguments: Try[(WdlValue, WdlValue)] = params.size match {
-      case 2 => (params.head |@| params.tail.head) map { (_, _) }
+      case 2 => (params.head, params.tail.head).mapN { (_, _) }
```
mcovarr: you will be missed, @helgridly operator.  😢

danbills:
@cjllanwarne  The aim here is implementing pre-processing and tying it into parsing.

A service can sit on top of this, no?  What I took away from the service registry in Cromwell is that it was an actor dispatching calls according to registered services.

cwl/src/test/resources/bad2.cwl
danbills: They are the same.  The test was originally intended to make sure it picked up 2 separate errors, as two errors re: `bad.cwl` could feasibly be the same error twice. 

I see now that I'm not asserting the 2 separate error condition.  I'll fix it up.

cwl/src/main/scala/wdl4s/cwl/CwlFile.scala
```diff
@@ -239,6 +137,18 @@ case class CommandLineTool(
...
+
+    val cwi = CallNode.callWithInputs(id.getOrElse("this is a made up call node name"), taskDefinition, Map.empty)
+
+    Set.empty[GraphNode] ++ cwi.inputs + cwi.call
+  }
+
+  def asCwl = Coproduct[Cwl](this)
 }
 
 object CommandLineTool {
```
danbills, line 153: * Move this to `CommandLineTool.scala` and get rid of this file.
 * Clean up CommandLineTool.scala.  Probably should create separate files for model objects in their own package...

cwl/src/main/scala/wdl4s/cwl/Workflow.scala
```diff
@@ -0,0 +1,134 @@
...
+      )
+    )
+  }
+
+  def asCwl = Coproduct[Cwl](this)
+}
+object Workflow {
+
+  //run this thing and give me an output
+  def minimal(wfs: WorkflowStep) = {
```
danbills: delete me

cwl/src/main/scala/wdl4s/cwl/Workflow.scala
```diff
@@ -0,0 +1,134 @@
...
+    inputs: Array[InputParameter] = Array.empty,
+    outputs: Array[WorkflowOutputParameter] = Array.empty): Either[NonEmptyList[String], Cwl] = {
+
+
+    //does no checking at all
+    val wf = Workflow(steps = steps.toArray)
+    Right(Coproduct[Cwl](wf))
+  }
+
+  val `class`: Witness.`"Workflow"`.T = "Workflow".narrow
```
danbills: * move this to Constructor type and get rid of this as it's not being properly enforced.
 * make Workflow constructor private and direct people to use constructor above, so that awkwardness around providing this value is avoided.

cwl/src/main/scala/wdl4s/cwl/Workflow.scala
```diff
@@ -0,0 +1,134 @@
...
+  }
+
+  def asCwl = Coproduct[Cwl](this)
+}
+object Workflow {
+
+  //run this thing and give me an output
+  def minimal(wfs: WorkflowStep) = {
+  }
+  def validate(
```
danbills: rename to `create` or something better

cwl/src/test/scala/wdl4s/cwl/CwlDecoderSpec.scala
```diff
@@ -0,0 +1,31 @@
...
+      }
+
+  property("broken links fail the SALAD preprocessing test") =
+    decodeAllCwl(rootPath/"brokenlinks.cwl").
+      value.
+      unsafeRunSync.
+      isLeft
+
+  property("links fail to parse breaks the SALAD preprocessing test") =
+    decodeAllCwl(rootPath/"links_dont_parse.cwl").
```
danbills, line 28: assert 2 separate errors here re: `bad.cwl` and `bad2.cwl`

cwl/src/test/scala/wdl4s/cwl/CwlWorkflowWomSpec.scala
```diff
@@ -25,94 +29,40 @@ class CwlWorkflowWomSpec extends FlatSpec with Matchers {
...
-    decodeCwl(firstWorkflow) map {
-      case (workflow: Workflow, nameToFile) =>
-        workflow.womExecutable(nameToFile) match {
-          case Valid(ex) => validateWom(ex)
-          case Invalid(errors) => fail(s"executable was not created." + errors.toList.mkString("\n", "\n", "\n"))
-        }
-      case (wth: Any, _) => fail(s"Parsed unexpected CwlFile subtype $wth")
-    } leftMap { error => fail(s"did not parse!  $error") }
+      ex <- wf.womExecutable.toEither
+    } yield validateWom(ex)
```
danbills: assert that a `Left` will report an error

cwl/src/test/scala/wdl4s/cwl/CwlWorkflowWomSpec.scala
```diff
@@ -121,151 +71,44 @@ name: "file:///home/dan/wdl4s/r.cwl"
...
-    val workflow = CwlCodecs.decodeCwl(threeStep) match {
-      case Valid((wf: Workflow, _)) => wf
-      case o => fail(s"didn't parse a workflow! $o")
-    }
-    val wf = workflow.womExecutable(Map.empty) match {
+    val wf = decodeAllCwl(rootPath/"three_step.cwl").map {
+      _.select[Workflow].get
+    }.value.unsafeRunSync.fold(error => throw new RuntimeException(s"broken parse! msg was $error"), identity)
+
+    val wfd = wf.womExecutable match {
```
danbills, line 105: these abbreviated names aren't doing it for  me

mcovarr (commented):
First tranche of comments, there will likely be more to come.  🙂

wom/src/main/scala/wdl4s/wdl/expression/WdlStandardLibraryFunctions.scala
cjllanwarne: Is that map really a `.tupled`

danbills:
yes

wom/src/main/scala/wdl4s/wdl/expression/WdlStandardLibraryFunctions.scala
danbills: yep

cwl/src/main/scala/wdl4s/cwl/AddEmbeddedCwl.scala
```diff
@@ -0,0 +1,50 @@
...
+  implicit def workflow =
+    at[Workflow] {
+      wf =>
+        EitherT {
+          wf.steps.toList.
+            flatMap(_.run.select[String].toList).
+            traverse[ParseValidated, (String, Cwl)] {
+              path =>
+                CwlDecoder.
+                  decodeAllCwl(Path(path.drop(5))).
```
mcovarr: what is `drop(5)` about?

cwl/src/main/scala/wdl4s/cwl/AddEmbeddedCwl.scala
```diff
@@ -0,0 +1,50 @@
...
+            map(_.toEither)
+        }.map{
+          map =>
+
+            workflowStepLens.modify(wf){
+              _.map{ step =>
+                workflowStepRunLens.modify(step)(_.fold(RunToEmbeddedCwl).apply(map))
+              }
+            }.asCwl
+        }
```
mcovarr: intermediate vals and comments throughout would be a big help

cwl/src/main/scala/wdl4s/cwl/AddEmbeddedCwl.scala
```diff
@@ -0,0 +1,50 @@
...
+import cats.{Applicative, Monad}
+import cats.data.EitherT._
+import cats.data.EitherT
+import lenthall.validation.ErrorOr._
+
+object AddEmbeddedCwl extends Poly1 {
+  val workflowStepLens = lens[Workflow].steps
+  val workflowStepRunLens = lens[WorkflowStep].run
+
+  implicit def workflow =
```
mcovarr: can there be return types on these `Poly1` `implicit`s?

cwl/src/test/scala/wdl4s/cwl/CwlWorkflowWomSpec.scala
```diff
@@ -25,94 +29,40 @@ class CwlWorkflowWomSpec extends FlatSpec with Matchers {
...
+    (for {
+      clt <- decodeAllCwl(rootPath/"1st-tool.cwl").
+              map(_.select[CommandLineTool].get).
+              value.
+              unsafeRunSync
+
+      wom <-  clt.womExecutable.toEither
+    } yield validateWom(wom)).leftMap(e => throw new RuntimeException(s"error! $e"))
+
+    def validateWom: Executable => Unit =
```
mcovarr: why `def` and not `val`?

mcovarr (commented):
another flyover

cwl/src/main/scala/wdl4s/cwl/AddEmbeddedCwl.scala
danbills: The actual return type is pretty gross and just as hard to understand :
image

cwl/src/main/scala/wdl4s/cwl/AddEmbeddedCwl.scala
danbills: Ammonite Path expects `/` as start character whereas SALAD outputs `file:///`.  An elegant solution wasn't immediately obvious so I dropped 5 characters.  I will write a comment explaining.

geoffjentry: Could this be done with the `better-files` stuff we're using elsewhere?

danbills: Yeah... that'd probably be better as it's consistent w/ other parts of the codebase and also doesn't require this `drop` shenanigan

cwl/src/main/scala/wdl4s/cwl/AddEmbeddedCwl.scala
mcovarr: What is even going on there?  Is the return type is followed by a block??

danbills: Exactly :).  `Result` type is being assigned within the body of this specific `Case`, and yeah TIL you can declare types w/ blocks

cwl/src/test/scala/wdl4s/cwl/CwlWorkflowWomSpec.scala
danbills: I believe they are equivalent in this case so no reason in particular.  If you want a `val` you shall have a `val`

mcovarr: I wasn't sure if the def "interned" the function or created a new one each time, but the val is definitely a one-time thing.  🙂

cwl/src/main/scala/wdl4s/cwl/AddEmbeddedCwl.scala
```diff
@@ -0,0 +1,59 @@
...
+
+        //Gather up all the filenames from the "run" section of workflow steps.
+        val fileNames =
+          workflow.
+            steps.
+            toList.
+            flatMap(_.run.select[String].toList)
+
+        //read the files, parse them, and put them in a Map
+        val cwlMap: IO[Either[NonEmptyList[String], Map[String, Cwl]]] =
```
mcovarr: ~~is that inner thing an `ErrorOr[Map[String, Cwl]]`?~~

Edit: no.

cwl/src/main/scala/wdl4s/cwl/Decoder.scala
```diff
@@ -0,0 +1,53 @@
...
+import better.files.{File => BFile}
+
+import scala.util.Try
+
+object CwlDecoder {
+
+  type Parse[A] = EitherT[IO, NonEmptyList[String], A]
+  type ParseValidated[A] = IO[ValidatedNel[String, A]]
+
+  def salad(path: BFile): Parse[String] = {
```
mcovarr: It's probably not clear to most readers that `salad` is being used a verb in the imperative mood; a different name might help.  🙂

cwl/src/main/scala/wdl4s/cwl/Decoder.scala
```diff
@@ -0,0 +1,53 @@
...
+        toEither.
+        leftMap(t => NonEmptyList.one(s"running cwltool on file ${path.toString} failed with ${t.getMessage}"))
+
+    EitherT {
+      IO {
+        for {
+          _cwlToolResult <- cwlToolResult
+          result <- resultToEither(_cwlToolResult)
+        } yield result
+      }
```
mcovarr: is this different from

```
    EitherT {
      IO {
        cwlToolResult flatMap resultToEither
      }
    }
```

danbills: Nope!  Thanks very much!

cwl/src/main/scala/wdl4s/cwl/Workflow.scala
danbills: created #184

wom/src/main/scala/wdl4s/wdl/WdlCall.scala
```diff
@@ -60,7 +60,7 @@ object WdlCall {
     }
 
     for {
-      combined <- (graphNodeInputExpressionValidation |@| wdlCall.callable.womDefinition).tupled
+      combined <- (graphNodeInputExpressionValidation, wdlCall.callable.womDefinition).tupled
```
cjllanwarne: `|'o'|`

mcovarr:
:+1:

[Approved with PullApprove](https://pullapprove.com/broadinstitute/wdl4s/pull-request/183/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=broadinstitute/wdl4s)

gemmalam:
@cjllanwarne can you re-review this? @danbills is waiting on one more 👍

cwl/src/test/scala/wdl4s/cwl/ParseBigThreeSpec.scala
```diff
@@ -2,81 +2,32 @@ package wdl4s.cwl
...
-outputs: []
-id: file:///Users/danb/common-workflow-language/v1.0/examples/env.cwl
-name: file:///Users/danb/common-workflow-language/v1.0/examples/env.cwl
-"""
-
-    decodeCwl(envCwl).isValid should be (true)
+    decodeAllCwl(rootPath/"env.cwl").
+      value.
+      unsafeRunSync.
+      isRight shouldBe true
```
cjllanwarne, line 31: ToL (depending on how likely you think it is that this might break): boolean assertions can be tedious to debug. I find it better to do a `match` and nicely messaged:
```
case Left(_) => // great
case Right(error) => fail("Didn't do 'X' because $error")
```

cjllanwarne (commented):
Before I give a thumb, I just want to double-check this is the right strategy:
- I don't quite understand how this fits into a "Salad as a service" world. I had anticipated the `SALAD`ing would have to happen via the service registry in Cromwell so that we can choose how best to scale it up. This way seems to hard-code the command line variant (and therefore all of the supporting dependencies) into `wdl4s`.

Other comments:
- Seems like there's at least two changes going on here (SALAD, cats++, ), my personal opinion is that small "fit in head" PRs are better than big world-changing "omnibus" PRs. This was mitigated a little bit in this case by...:
- Mini-confession: I totally skimmed over the `Coproduct`y stuff in `wdl4s.cwl` because I don't have much context for it and don't really know what effect the changes are having.

cjllanwarne:
One other comment, there seem to be a few notes-to-self still in the  comment history. Could you confirm they've been done (or won't be done in this PR)?

cwl/src/test/scala/wdl4s/cwl/ParseBigThreeSpec.scala
danbills: 👍 I'll make these assertions more meaningful

danbills:
Discussed outside of this PR.  Summarizing:

Salad as a service:
 * This can be achieved via a service registry above this implementation, similar to the Metadata Service.  We do not incur too much dependency pain as we are importing Ammonite to run sub-processes and nothing SALAD-specific.
 * This seems tightly bound to WDL4s CWL package.  This is on purpose.  Our CWL model is making the assumption that this pre-processing has taken place, which allows us to simplify our approach.  Any user of this library would need such a pre-processing implementation or risk losing functionality. 
 

General lessons learned are:
 * Smaller, more-focused PR are easier to review and have less noise.
 * Separate PR's to organize/refactor code should be done in lieu of "while I'm here I'll do this"

Moving forward, I will take another look at my notes-to-self and either
 * fix them
 * make meaningful comments as to their status
 * create an issue

cjllanwarne:
👍 once the notes to self have been verified

[Approved with PullApprove](https://pullapprove.com/broadinstitute/wdl4s/pull-request/183/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=broadinstitute/wdl4s)

cwl/src/main/scala/wdl4s/cwl/CwlFile.scala
danbills: created #185 so as not to further dilute this PR

cwl/src/test/scala/wdl4s/cwl/ParseBigThreeSpec.scala
danbills: The message is provided in the description of the test, and thus writing out an error becomes essentially  `s"did not $description"`.  I think with a test consisting of a single assertion this layout is pretty reasonable as-is.

PR #187, opened 2017-08-31
Fix cross publish

cjllanwarne:


cjllanwarne:
I'm going to fast track this PR without reviews.

PR #211, opened 2017-09-14
Prevent the witness from sometimes committing perjury in cwl.CommandL…

curoli:
Prevent the witness from sometimes committing perjury in cwl.CommandLineTool, the same way it was done before for cwl.Workflow.

### DIFF
--- a/cwl/src/main/scala/wdl4s/cwl/CwlFile.scala
+++ b/cwl/src/main/scala/wdl4s/cwl/CwlFile.scala
@@ -1,11 +1,9 @@
 package wdl4s.cwl
 
-import shapeless.syntax.singleton._
 import shapeless._
 import cats.syntax.foldable._
 import shapeless.{:+:, CNil, Poly1, Witness}
 import CwlType._
-import shapeless.syntax.singleton._
 import CwlVersion._
 import cats.data.Validated._
 import lenthall.validation.ErrorOr._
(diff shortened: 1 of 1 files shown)


danbills:
Please attempt a unit test that shows that this enforces CommandLineTool's `class` to always be "CommandLineTool."

I don't think it's possible, as it's not enforced when declared this way. 

I'm looking at your loam stream repo to try to figure out what is going on.

curoli:
I don't know how to make Unit tests about type safety, because I can't put stuff into unit tests that doesn't compile.

But let's see:

CommandLineTool()  //  compiles - good
CommandLineTool(`class` = "Hello")  //  does not compile - good
CommandLineTool(`class` = "Hello".narrow)  //  does not compile - good
CommandLineTool(`class` = "CommandLineTool")  //  does not compile - I think that's fine
CommandLineTool(`class` = "CommandLineTool",narrow)  //  does not compile - I think that's fine
CommandLineTool(`class` = CommandLineTool.`class`)  //  compiles - good
CommandLineTool(`class` = Workflow.`class`)  //  does not compile - good

That "CommandLineTool" and "CommandLineTool".narrow might look less than ideal, but I think it's OK, because using CommandLineTool.`class` is better anyway. Besides, Shapeless calls these things singletons, so by design, you're not supposed to create a second instance of it, right?

With asInstanceOf[ClassType], you can assign anything, of course, but that's by design of asInstanceOf.

The remaining problem is that the Shapeless Witness is broken. The same code sometimes works in one place but not in another for no apparent reason. No amount of clever coding or testing can make us safe until Shapeless fixes it.

danbills:
closing this as fix was applied in #214

PR #218, opened 2017-09-18
Wom Input Mappings

Horneth:
A few things here:

- The call node provides mappings from input definition to output port so that inputs can be evaluated at runtime by the engine.
- WDL currently does not allow overriding of *declared (e.g: `String my_decl = "hey"`* inputs by workflow inputs. Account for that by NOT creating a GraphInputNode for those declarations (they are not graph inputs). This is true both for task declarations and workflow declarations.
- Adds a method on `Graph` that takes a `Map[FQN, Any]` and validates that all graph inputs are satisfied, and returns a `Map[OutputPort, ResolvedWorkflowInput]`. A resolved workflow input is either a WdlValue or a `WomExpression` (if no input was provided and we fell back to the default expression value). Note that if we want default values to be able to reference other nodes, we should somehow instantiate this `WomExpression` to get an `InstantiatedExpression`.


cwl/src/test/scala/wdl4s/cwl/CwlWorkflowWomSpec.scala
```diff
@@ -128,6 +131,17 @@ class CwlWorkflowWomSpec extends FlatSpec with Matchers {
...
 
     wc.upstream.filter(_ eq ps).size shouldBe 1
+    
+    // Check that the inputDefinitionMappings are correct
+    ps.inputDefinitionMappings shouldBe empty
+    cgrep.inputDefinitionMappings should have size 2
+    
+    val cgrepFileInputDef = cgrep.callable.inputs.find(_.name == "file").get
+    val psStdoutOutputPort = ps.outputPorts.head
+    cgrep.inputDefinitionMappings(cgrepFileInputDef) shouldBe psStdoutOutputPort
```
mcovarr: are these instance equality checks? should they be?

wom/src/main/scala/wdl4s/wom/graph/CallNode.scala
```diff
@@ -83,18 +94,40 @@ object CallNode {
...
+        instantiatedExpressionInputs.get(inputDefinition.name) map { expr =>
+          val outputPort = InstantiatedExpressionNode(inputDefinition.name, expr).singleExpressionOutputPort
+          fold.copy(mappings = fold.mappings + (inputDefinition -> outputPort))
+        } getOrElse {
+          // Otherwise use the inputPorLinker
+          val linked = inputPortLinker(inputDefinition)
+          fold.copy(
+            mappings = fold.mappings + (inputDefinition -> linked.outputPort),
+            linkedInputPorts = fold.linkedInputPorts + linked
+          )
```
mcovarr: a `match` might be more clear here?

wom/src/main/scala/wdl4s/wom/graph/CallNode.scala
```diff
@@ -83,18 +94,40 @@ object CallNode {
...
+        * new input node. The inputPortLinker is the one that checks for an existing outputPort before creating the 
+        * input node if necessary.
+       */
+      def foldFunction(fold: InputDefinitionFold, inputDefinition: InputDefinition): InputDefinitionFold = {
+        // Check for an expression
+        instantiatedExpressionInputs.get(inputDefinition.name) map { expr =>
+          val outputPort = InstantiatedExpressionNode(inputDefinition.name, expr).singleExpressionOutputPort
+          fold.copy(mappings = fold.mappings + (inputDefinition -> outputPort))
+        } getOrElse {
+          // Otherwise use the inputPorLinker
```
mcovarr: inputPorLinker => inputPortLinker

wom/src/main/scala/wdl4s/wom/graph/CallNode.scala
```diff
@@ -83,18 +94,40 @@ object CallNode {
 
     instantiatedExpressionInputsAttempt map { instantiatedExpressionInputs =>
       val inputPortLinker = GraphNode.linkInputPort(name + prefixSeparator, portInputs, graphNodeSetter.get) _
+      
+      /*
+        * This is a bit misleading as it makes it seem like expressions have precedence over output ports. They don't.
+        * What matters is that we look for an expression OR a pre-existing output port before we generate a 
+        * new input node. The inputPortLinker is the one that checks for an existing outputPort before creating the 
+        * input node if necessary.
```
mcovarr: but there's no point checking for an output port if there's an expression?

wom/src/main/scala/wdl4s/wom/graph/Graph.scala
```diff
@@ -24,10 +28,41 @@ final case class Graph private(nodes: Set[GraphNode]) {
...
+      case gin: ExternalGraphInputNode => 
+        // The compiler needs the type ascription for some reason
+        (gin.singleOutputPort: OutputPort) -> fromInputMapping(gin).getOrElse(fallBack(gin))
+    }).toMap.sequence
+  }
 }
 
 object Graph {
 
+  type ResolvedWorkflowInput = Either[WdlValue, WomExpression]
```
mcovarr: ToL I know we've talked about two type coproducts versus `Either` before, but seeing these `Left.apply`s and `Right`s is kind of jarring.

wom/src/test/scala/wdl4s/wom/graph/CallNodeSpec.scala
```diff
@@ -0,0 +1,123 @@
...
+import wdl4s.wom.RuntimeAttributes
+import wdl4s.wom.callable.Callable.{DeclaredInputDefinition, OptionalInputDefinition, OptionalInputDefinitionWithDefault, RequiredInputDefinition}
+import wdl4s.wom.callable.TaskDefinition
+import wdl4s.wom.expression.PlaceholderWomExpression
+import wdl4s.wom.graph.GraphNodePort.GraphNodeOutputPort
+
+class CallNodeSpec extends FlatSpec with Matchers {
+  behavior of "CallNode"
+
+  def ginNotFound(name: String) = fail(s"Can't find an GraphInputNode for $name")
```
mcovarr: an => a

wom/src/test/scala/wdl4s/wom/graph/CallNodeSpec.scala
```diff
@@ -0,0 +1,123 @@
...
+      inputs = List(
+        requiredSatisfiedByPort, requiredSatisfiedByExpression, requiredUnsatisfied,
+        optionalWithDefaultSatisfiedByPort, optionalWithDefaultSatisfiedByExpression, optionalWithDefaultUnsatisfied,
+        optionalSatisfiedByPort, optionalSatisfiedByExpression, optionalUnsatisfied,
+        declarationSatisfiedByPort, declarationSatisfiedByExpression, declarationUnsatisfied
+      )
+    )
+
+    val requiredPort = GraphNodeOutputPort("required_port", WdlStringType, null)
+    val optionalPort = GraphNodeOutputPort("optiona_port", WdlStringType, null)
```
mcovarr: optiona_port => optional_port

wom/src/test/scala/wdl4s/wom/graph/CallNodeSpec.scala
```diff
@@ -0,0 +1,123 @@
...
+    // unsatisfied input defs should map to the above newInput output ports
+    inputDefinitionMappings(requiredUnsatisfied) eq requiredInput.singleOutputPort shouldBe true
+    inputDefinitionMappings(optionalUnsatisfied) eq optionalInput.singleOutputPort shouldBe true
+    inputDefinitionMappings(optionalWithDefaultUnsatisfied) eq optionalInputWithDefault.singleOutputPort shouldBe true
+    
+    // unsatisfied declared input def gets a FloatingExpressionNode
+    inputDefinitionMappings(declarationUnsatisfied).graphNode.asInstanceOf[UnInstantiatedExpressionNode].womExpression shouldBe declarationUnsatisfied.expression
+
+    /* *** Satisfied by port *** */
+    // satisfied by port should map to their respective output port from the
```
mcovarr: comment trails

mcovarr (commented):
first pass

cwl/src/test/scala/wdl4s/cwl/CwlWorkflowWomSpec.scala
```diff
@@ -109,7 +109,10 @@ class CwlWorkflowWomSpec extends FlatSpec with Matchers {
 
     val nodes = wfd.innerGraph.nodes
 
-    nodes collect { case gin: GraphInputNode => gin.name } should be(Set("file:///Users/danb/wdl4s/r.cwl#pattern"))
+    val graphInputNodes = nodes collect { case gin: GraphInputNode => gin }
+    graphInputNodes should have size 1
+    val patternInputNode = graphInputNodes.head
+    patternInputNode.name should be("file:///Users/danb/wdl4s/r.cwl#pattern")
```
mcovarr, line 116: why these changes? this looks essentially equivalent to the original on the happy path.

wom/src/main/scala/wdl4s/wom/graph/CallNode.scala
Horneth: No, basically WDL uses expressions and CWL could use either port or expression, but not both at the same time. So in practice we should only ever have one of them.

wom/src/main/scala/wdl4s/wom/graph/Graph.scala
Horneth: Ok I'll give it a shot

cwl/src/test/scala/wdl4s/cwl/CwlWorkflowWomSpec.scala
Horneth: hmm not sure if they are but they definitely should

wom/src/main/scala/wdl4s/wom/callable/Callable.scala
```diff
@@ -20,9 +20,18 @@ object Callable {
...
-  final case class OptionalInputDefinitionWithDefault(name: String, womType: WdlType, default: WomExpression) extends InputDefinition
-  final case class RequiredInputDefinition(name: String, womType: WdlType) extends InputDefinition
+  /**
+    * InputDefinitions that translate to GraphInputNodes when left unconnected
+    */
+  sealed trait GraphInputDefinition extends InputDefinition
 
-  case class OutputDefinition(name: String, womType: WdlType, expression: WomExpression)
+  final case class OptionalInputDefinition(name: String, womType: WdlOptionalType) extends GraphInputDefinition
+  final case class OptionalInputDefinitionWithDefault(name: String, womType: WdlType, default: WomExpression) extends GraphInputDefinition
```
mcovarr, line 29: woah...a non-optional that's assigned a value cannot be overridden by a workflow input, but an optional assigned a default can?

wom/src/main/scala/wdl4s/wom/graph/CallNode.scala
mcovarr: ok cool, just wanted to confirm 🙂

wom/src/main/scala/wdl4s/wom/graph/GraphNode.scala
```diff
@@ -77,20 +79,28 @@ object GraphNode {
...
-      case input if inputMapping.contains(input.name) => (inputMapping(input.name), None)
+
+    def withInputPort(outputPort: OutputPort, graphNode: Option[GraphInputNode]) = {
+      val inputPort = ConnectedInputPort(inputDefinition.name, inputDefinition.womType, outputPort, callNodeRef)
+      LinkedInputPort(Option(inputPort), outputPort, graphNode)
+    }
+
+    // Unsatisfied GraphInputDefinitions generate ExternalGraphInputNodes
+    inputDefinition match {
+      case input if inputMapping.contains(input.name) => withInputPort(inputMapping(input.name), None)
```
mcovarr: dumb question: what is this case?

wom/src/test/scala/wdl4s/wdl/wom/WdlScatterWomSpec.scala
```diff
@@ -176,7 +176,7 @@ class WdlScatterWomSpec extends FlatSpec with Matchers {
 
       // Find the inputs:
       val inputNodes: Set[GraphInputNode] = workflowGraph.nodes.filterByType[GraphInputNode]
-      inputNodes.map {_.name} should be(Set("x", "foo.j"))
+      inputNodes.map {_.name} should be(Set("foo.j"))
```
mcovarr, line 177: Just checking: this change is because `x` is a `DeclaredInputDefinition` and no longer generates a `GraphInputNode`?

wom/src/main/scala/wdl4s/wom/graph/UnInstantiatedExpressionNode.scala
```diff
@@ -0,0 +1,30 @@
...
+
+import wdl4s.wdl.types.WdlType
+import wdl4s.wom.expression.WomExpression
+import wdl4s.wom.graph.GraphNodePort.{GraphNodeOutputPort, InputPort}
+
+
+/**
+  * Used to hold an un-instantiated expression. This is useful for a task declaration for example
+  * for which expression inputs are not explicitly mapped to other output ports but are instead derived from preceding
+  * inputs.
```
mcovarr: It feels a little weird that we need this as a separate concept. Might it make sense to have a lazy evaluation that could be used in this case to yield output ports for preceding inputs? Or maybe just a name that better reflects why this concept exists and not just that it means the expression isn't instantiated.

wom/src/test/scala/wdl4s/wom/graph/CallNodeSpec.scala
```diff
@@ -0,0 +1,123 @@
...
+    val inputDefinitionMappings = callWithInputs.node.inputDefinitionMappings
+    inputDefinitionMappings.size == taskDefinition.inputs.size shouldBe true
+
+    /* *** Unsatisfied *** */
+    // unsatisfied input defs should map to the above newInput output ports
+    inputDefinitionMappings(requiredUnsatisfied) eq requiredInput.singleOutputPort shouldBe true
+    inputDefinitionMappings(optionalUnsatisfied) eq optionalInput.singleOutputPort shouldBe true
+    inputDefinitionMappings(optionalWithDefaultUnsatisfied) eq optionalInputWithDefault.singleOutputPort shouldBe true
+    
+    // unsatisfied declared input def gets a FloatingExpressionNode
```
mcovarr: FloatingExpressionNode!

mcovarr (commented):
second pass

wom/src/test/scala/wdl4s/wom/graph/CallNodeSpec.scala
Horneth: argh I renamed the class but not the comment

wom/src/main/scala/wdl4s/wom/graph/UnInstantiatedExpressionNode.scala
Horneth: Yeah the reason for that is that I wanted to have a `Map[InputDefinition, OutputPort]` rather than a `Map[InputDefinition, Either[OutputPort, WomExpression]]` or 2 maps.
And to have an OutputPort one needs a GraphNode to hold it, hence this class.

But I guess we could have 2 maps or one with a complex value type and then we wouldn't need this.

wom/src/main/scala/wdl4s/wom/graph/CallNode.scala
Horneth: Hmm do you mean like
```
          fold match {
            case InputDefinitionFold(mappings, linkedInputPorts) =>
              fold.copy(
                mappings = mappings + (inputDefinition -> linked.outputPort),
                linkedInputPorts = linkedInputPorts + linked
              )
          }
```
?

mcovarr: Oh, sorry that wasn't clear.  I meant:

```
instantiatedExpressionInputs.get(inputDefinition.name) match {
  case Some(expr) => // expression
  case None => // use input port linker
} 
```

Horneth: oh right yep that looks better

wom/src/main/scala/wdl4s/wom/callable/Callable.scala
Horneth: Well it depends, in WDL this is what happens

```
task t {
    String t1 = "t1" # final value: "t1"
    String t2 = "t2" # final value: "overridden"
}

workflow w {
    String w1 = "w1" # final value "w1"
    
    call t { input: t2 = "overridden" }
}
```

```
{
  "w.w1: "not_overridden",
  "w.t.t1: "not_overridden"
}
```

However CWL considers defaults to be real defaults and therefore the workflow inputs take precedence over the default value..

wom/src/test/scala/wdl4s/wdl/wom/WdlScatterWomSpec.scala
Horneth: Yep

wom/src/main/scala/wdl4s/wom/graph/GraphNode.scala
Horneth: I don't think this can happen from a WDL workflow since everything is an expression but in CWL, before calling `callWithInputs` we actively try to map all inputs of the step to other step outputs or workflow inputs. So we actually expect the outputPort to be in this `inputMapping` most of the time. If it isn't then we need a new GraphInputNode.

cwl/src/test/scala/wdl4s/cwl/CwlWorkflowWomSpec.scala
Horneth: Because there's an equality check using `patternInputNode` at the end so I needed access to it.
I don't see a reason for graphInputNodes though I can inline that.

Horneth: Actually `graphInputNodes` is used both to check its size and get its head

Horneth:
@mcovarr I think this is ready for pass 3 :)

wom/src/main/scala/wdl4s/wdl/Scope.scala
```diff
@@ -36,7 +38,15 @@ trait Scope {
   }
 
   lazy val childGraphNodes: Set[WdlGraphNode] = children.toSet.filterByType[WdlGraphNode]
-  lazy val childGraphNodesSorted: List[WdlGraphNode] = childGraphNodes.toList.sortWith((first, second) => first.isUpstreamFrom(second))
+  lazy val childGraphNodesSorted: List[WdlGraphNode] = {
+    // We can't use a classic ordering because the upstream / downstream order is not total over the set of nodes
+    // Instead we use topological sorting to guarantee that we process the nodes top to bottom
+    val edges = childGraphNodes flatMap { child => child.upstreamAncestry map { upstreamNode => DiEdge(upstreamNode, child) } }
```
Horneth, line 44: We might need to talk about this

cwl/src/test/scala/wdl4s/cwl/CwlWorkflowWomSpec.scala
mcovarr: This is all ToL but: yes, on the unhappy path. And if the sets mismatched it should be clear why. Anyway, doesn't matter.

wom/src/main/scala/wdl4s/wdl/Scope.scala
mcovarr: 😬

wom/src/main/scala/wdl4s/wom/graph/Graph.scala
```diff
@@ -1,14 +1,19 @@
...
 import cats.instances.list._
+import cats.syntax.apply._
 import cats.syntax.functor._
 import cats.syntax.traverse._
-
-import lenthall.validation.ErrorOr.ErrorOr
+import cats.syntax.validated._
 import lenthall.collections.EnhancedCollections._
-import wdl4s.wom.graph.GraphNodePort.InputPort
+import lenthall.validation.ErrorOr.{ErrorOr, _}
```
kshakir, line 9: `import lenthall.validation.ErrorOr._`

wom/src/test/scala/wdl4s/wdl/wom/WdlWomExpressionsAsInputsSpec.scala
```diff
@@ -62,9 +62,13 @@ class WdlWomExpressionsAsInputsSpec extends FlatSpec with Matchers {
...
 
+    val a = callNodes("a")
+    val b = callNodes("b")
     val c = callNodes("c")
-    c.expressionBasedInputs should have size 1
-    val inputExpression = c.expressionBasedInputs.values.head
+    c.inputPorts should have size 2
+    c.inputPorts.map(_.upstream) shouldBe a.outputPorts ++ b.outputPorts
+
+    val inputExpression = c.inputDefinitionMappings.head._2.select[InstantiatedExpression].get
```
kshakir, line 71: `.get` => `.getOrElse(fail(...))`

kshakir (approved):
A couple minor things

wom/src/main/scala/wdl4s/wom/graph/CallNode.scala
```diff
@@ -84,17 +101,107 @@ object CallNode {
...
-      // Filter out the inputs we already have from expressions:
-      val asYetUnsuppliedInputs = callable.inputs.filterNot(inputDef => instantiatedExpressionInputs.contains(inputDef.name))
-      val linkedInputPortsAndGraphInputNodes = asYetUnsuppliedInputs map inputPortLinker
-
-      val linkedInputPorts = linkedInputPortsAndGraphInputNodes.map(_.newInputPort)
-      val graphInputNodes = linkedInputPortsAndGraphInputNodes collect { case LinkedInputPort(_, Some(gin)) => gin }
+      /*
+        * This method is used when we didn't find an instantiated expression for a DeclaredInputDefinition,
+        * but there still might be a non external output port that matches 
+        * (remember we don't want to to override DeclaredInputDefinition with external inputs).
```
mcovarr, line 107: to to => to

mcovarr: I don't understand what's going on here.

Horneth:
Closing this after discussion with @mcovarr and @cjllanwarne yesterday. I'll re-open when ready

PR #221, opened 2017-09-18
Make CWL Womification more robust

danbills:
There is quite a bit of technical debt in CWL -> WOM.

This PR addresses the getOrElse that took a pure value from an `ErrorOr`.

As we were `flatMap`-ing quite regularly, I took the initiative to introduce the type parameter:
`type Checked[A] = Either[NonEmptyList[String], A]` which is analogous to `ErrorOr`.

### DIFF
--- a/cwl/src/main/scala/wdl4s/cwl/CwlFile.scala
+++ b/cwl/src/main/scala/wdl4s/cwl/CwlFile.scala
@@ -1,7 +1,5 @@
 package wdl4s.cwl
 
-import cats.data.Validated._
-import lenthall.validation.ErrorOr._
 import shapeless.syntax.singleton._
 import shapeless.{:+:, CNil, Poly1, Witness, _}
 import wdl4s.cwl.CommandLineTool.{BaseCommand, StringOrExpression}
@@ -56,8 +54,8 @@ case class CommandLineTool private(
                                    temporaryFailCodes: Option[Array[Int]],
                                    permanentFailCodes: Option[Array[Int]]) {
 
-  def womExecutable: ErrorOr[Executable] =
-    Valid(Executable(taskDefinition))
+  def womExecutable: Checked[Executable] =
+    Right(Executable(taskDefinition))
 
 
   object BaseCommandToString extends Poly1 {
--- a/cwl/src/main/scala/wdl4s/cwl/CwlFile.scala
+++ b/cwl/src/main/scala/wdl4s/cwl/CwlFile.scala
@@ -10,6 +10,7 @@ import wdl4s.wom.callable.{Callable, TaskDefinition}
 import wdl4s.wom.executable.Executable
 import wdl4s.wom.expression.WomExpression
 import wdl4s.wom.{CommandPart, RuntimeAttributes}
+import lenthall.Checked
 
 /**
   *
@@ -104,7 +105,7 @@ case class CommandLineTool private(
     //TODO: This output does _not_ capture expressions from the output.outputBinding
     //The implementation must include the expression evaluation pieces as detailed in:
     //http://www.commonwl.org/v1.0/CommandLineTool.html#CommandOutputBinding
-    
+
     // For inputs and outputs, we only keep the variable name in the definition
     val outputs: Set[Callable.OutputDefinition] = this.outputs.map {
       output =>
--- a/cwl/src/main/scala/wdl4s/cwl/package.scala
+++ b/cwl/src/main/scala/wdl4s/cwl/package.scala
@@ -1,9 +1,9 @@
 package wdl4s
 
+import cats.data.NonEmptyList
 import eu.timepit.refined.api.Refined
 import eu.timepit.refined.string._
 import eu.timepit.refined._
-import lenthall.validation.ErrorOr.ErrorOr
 import wdl4s.cwl.CwlType.CwlType
 import wdl4s.cwl.CwlType._
 import wdl4s.wdl.types._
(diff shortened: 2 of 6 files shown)


cwl/src/main/scala/wdl4s/cwl/CwlFile.scala
```diff
@@ -56,8 +54,8 @@ case class CommandLineTool private(
                                    temporaryFailCodes: Option[Array[Int]],
                                    permanentFailCodes: Option[Array[Int]]) {
 
-  def womExecutable: ErrorOr[Executable] =
-    Valid(Executable(taskDefinition))
+  def womExecutable: Checked[Executable] =
+    Right(Executable(taskDefinition))
```
cjllanwarne, line 59: ToL: any opinions on typedef-ing `Right` and `Left` to have more `Checked`-like names too? eg `Left` could be `Wrong`? Just a thought...

cwl/src/main/scala/wdl4s/cwl/package.scala
```diff
@@ -52,6 +33,8 @@ package object cwl extends TypeAliases {
 
   type Cwl = Workflow :+: CommandLineTool :+: CNil
 
+  type Checked[A] = Either[NonEmptyList[String], A]
```
cjllanwarne: I think it'd be cool if this was added in `lenthall` alongside some helper extensions (eg `toErrorOr`?) if we can start using it elsewhere too?

cwl/src/test/scala/wdl4s/cwl/CwlWorkflowWomSpec.scala
```diff
@@ -49,7 +48,7 @@ class CwlWorkflowWomSpec extends FlatSpec with Matchers {
               map(_.select[CommandLineTool].get).
               value.
               unsafeRunSync
-      wom <-  clt.womExecutable.toEither
+      wom <-  clt.womExecutable
```
cjllanwarne: Extra space persists

cwl/src/test/scala/wdl4s/cwl/CwlWorkflowWomSpec.scala
```diff
@@ -103,7 +102,7 @@ class CwlWorkflowWomSpec extends FlatSpec with Matchers {
     }.value.unsafeRunSync.fold(error => throw new RuntimeException(s"broken parse! msg was $error"), identity)
 
     val wfd = wf.womExecutable match {
-      case Valid(Executable(wf: WorkflowDefinition)) => wf
+      case Right(Executable(wf: WorkflowDefinition)) => wf
       case o => fail(s"invalid executable $o")
```
cjllanwarne: NYF: This case looks like it'll print out `invalid executable Left(blah)`, instead of just `invalid executable blah`

cjllanwarne (approved):
Small comments, including a preference to put `Checked` into `lenthall` if you think it might be useful elsewhere too

Horneth (approved):
@danbills 
You probably explained that already but I either missed it or forgot sorry :/
What's the value of `Checked` over `ErrorOr` ?

cwl/src/main/scala/wdl4s/cwl/CwlFile.scala
danbills: I believe Either has the proper semantics of "correct" and "wrong", no?  After all it is right-biased.

cwl/src/main/scala/wdl4s/cwl/package.scala
danbills: Yes it should be.  I'll create a PR

cwl/src/test/scala/wdl4s/cwl/CwlWorkflowWomSpec.scala
danbills: ah shoot!  This CWL subproject is going to be the first to get auto-formatted, so stand back!

cwl/src/test/scala/wdl4s/cwl/CwlWorkflowWomSpec.scala
danbills: Indeed!  I'm pretty sure this is endemic to my tests.

cwl/src/main/scala/wdl4s/cwl/CwlFile.scala
geoffjentry: FWIW I've seen libraries do something similar (Venners' `Bad` and `Good` come to mind) and I don't really see the point. If anything they often worry me about being *too* heavy handed (what if something is objectively "bad" but it's the thing I want to survive the right bias?). I'd prefer to stick to Left/Right.

danbills:
@Horneth 
The way we use them, they are exactly the same.  There's a small theoretical distinction, and that's that `Validated` is an Applicative, but not a Monad.*

`Either` can be stacked/combined with other monads via `EitherT`.  There exists no such class for `Validated`.

For any instance where we want Validated behavior Applicative (i.e. `traverse`/`sequence`), it's easy to convert to/from.

Thus I would prefer to write new code using `Either` and only use `Validated` where necessary.

\*
This is because the Applicative [`map`is derived from Monad's `flatMap`](https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/Monad.scala#L15 ).  If there were a `flatMap` available for `Validated`, it would then take only the first error that occurred, instead of gathering all the errors up.  

Since we are using cats' `Validated` Applicative and our own hand-rolled `flatMap`, we avoid the wrong situation where the Applicative take the first `Invalid` case.

PR #222, opened 2017-09-19
Skip superfluous and error-prone variable reference check.

mcovarr:
Addresses the issue found in [this forum post](https://gatkforums.broadinstitute.org/wdl/discussion/10361/workflow-validates-runs-when-submitted-alone-fails-validation-doesnt-run-when-used-as-a-subworkflow#latest).

### DIFF
--- a/wom/src/main/scala/wdl4s/wdl/WdlNamespace.scala
+++ b/wom/src/main/scala/wdl4s/wdl/WdlNamespace.scala
@@ -339,10 +339,22 @@ object WdlNamespace {
         )
     }
 
+    /*
+     * Used to filter task declaration nodes whose parents are subworkflow calls. Subworkflows will cause task declarations to appear twice among
+     * the namespace's descendants: once with a parent that is the containing static workflow and again with a parent that is the dynamic workflow
+     * call. Looking for variable references when the parent is a workflow call will fail if the variable references a call output since only
+     * declarations are made children of workflow calls. This variable reference check is also redundant to the check that will happen with the
+     * copy of the declaration that has a static workflow parent.
+     */
+    def parentIsNotAWorkflowCall(declaration: DeclarationInterface): Boolean = declaration.parent match {
+      case Some(_: WdlWorkflowCall) => false
+      case _ => true
+    }
+
     val declarationErrors = for {
       descendant <- namespace.descendants
       declaration <- getDecls(descendant)
-      error <- validateDeclaration(declaration, wdlSyntaxErrorFormatter)
+      error <- validateDeclaration(declaration, wdlSyntaxErrorFormatter) if parentIsNotAWorkflowCall(declaration)
     } yield error
 
     val scatterErrors = for {
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -5,7 +5,7 @@ object Dependencies {
 
   val sprayJsonV = "1.3.2"
   val circeVersion = "0.9.0-M1"
-  val lenthallV = "0.28-f5f7f86-SNAP"
+  val lenthallV = "0.28-9d9a747-SNAP"
 
   // Internal collections of dependencies
 


geoffjentry:
👍 

[Approved with PullApprove](https://pullapprove.com/broadinstitute/wdl4s/pull-request/222/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=broadinstitute/wdl4s)

wom/src/main/scala/wdl4s/wdl/WdlNamespace.scala
```diff
@@ -419,9 +419,16 @@ object WdlNamespace {
     expression.variableReferences collect { case variable if container.resolveVariable(variable.terminal.sourceString).isEmpty => variable.terminal }
 
   private def validateDeclaration(declaration: DeclarationInterface, wdlSyntaxErrorFormatter: WdlSyntaxErrorFormatter): Seq[SyntaxError] = {
+    // Used to filter task declaration nodes whose parents are subworkflow calls. Subworkflows will cause task declarations to appear twice among
+    // the namespace's descendants: once with a parent that is the containing static workflow and again with a parent that is the dynamic workflow
+    // call. Looking for variable references when the parent is a workflow call will fail if the variable references a call output since only
+    // declarations are made children of workflow calls. This variable reference check is also redundant to the check that will happen with the
+    // copy of the declaration that has a static workflow parent.
```
cjllanwarne: This look like it should be a single javadoc style comment since it should be considered attached to the `parentIsNotAWorkflowCall ` method

wom/src/main/scala/wdl4s/wdl/WdlNamespace.scala
```diff
@@ -419,9 +419,16 @@ object WdlNamespace {
...
+    def parentIsNotAWorkflowCall(declaration: DeclarationInterface): Boolean = declaration.parent.collect({ case p: WdlWorkflowCall => p }).isEmpty
```
cjllanwarne: `filterByType[WdlWorkflowCall]`

wom/src/main/scala/wdl4s/wdl/WdlNamespace.scala
```diff
@@ -419,9 +419,16 @@ object WdlNamespace {
...
+
     val invalidVariableReferences = for {
       expr <- declaration.expression.toSeq
-      variable <- referencesToAbsentValues(declaration, expr)
+      variable <- referencesToAbsentValues(declaration, expr) if parentIsNotAWorkflowCall(declaration)
```
cjllanwarne: Could you do this check on line 344 (before we even call `validateDeclaration`) to avoid wasting the effort of re-validating any statically-validated declarations?

wom/src/main/scala/wdl4s/wdl/WdlNamespace.scala
mcovarr: This is an internal function and wouldn't render as Scaladoc, but I could certainly use the multi-line syntax if that was deemed more palatable.

wom/src/main/scala/wdl4s/wdl/WdlNamespace.scala
mcovarr: Looks like it, I can try.

wom/src/main/scala/wdl4s/wdl/WdlNamespace.scala
cjllanwarne: Re previous conversation, since it's an `Option` I think I'd lean towards a `match` here after all

wom/src/main/scala/wdl4s/wdl/WdlNamespace.scala
geoffjentry: Despite my previous thumb I'll chime in to say that I concur in that I prefer 

     /*
        foo
     */

for multiline comments. I realize that wasn't a multiline comment :)

mcovarr:
@cjllanwarne ready for final review.

Horneth (approved):
Thanks for fixing this ! :)
Could we have a test for it too ? A wdl4s unit test would be enough I think.

mcovarr:
@Horneth I created a [Centaur test](https://github.com/broadinstitute/centaur/pull/233) for this.

PR #223, opened 2017-09-21
Call Inputs v3

Horneth:
New attempt at call node creation, this time most of the logic is handled by language specific code.
This generates a lot of boilerplate code when creating nodes from WOM directly which shows in the tests.

PR #225, opened 2017-09-21
implement globbing for WDL/WOM

danbills:
We've introduced a method on Wom's `Callable`:

`callWithInputs` that takes the parameters required to build the wom callable.

Under the hood, the TaskDefinition has the ability to glob files and should propagate that ability to be used in cromwell. 

Further, this globbing ability depends on the language.  So we now require the caller to TaskDefinition to provide the implementation of `globFiles`, which is defined as `Map[String, WdlValue] => Set[WdlGlobFile]`.


danbills:
Adding do not merge as it makes sense to pull in @Horneth 's changes before this and re-basing.

wom/src/main/scala/wdl4s/wom/callable/TaskDefinition.scala
```diff
@@ -18,8 +20,11 @@ case class TaskDefinition(name: String,
                           parameterMeta: Map[String, String],
                           outputs: Set[Callable.OutputDefinition],
                           inputs: List[_ <: Callable.InputDefinition],
+                         //TODO: Move this out?  Not being used.  Or add prefix separator field on Callable
```
Horneth, line 23: This will go away with https://github.com/broadinstitute/wdl4s/pull/223

Horneth (commented):
Taking a second look, it seems to me that the only WDL specific bit that is actually passed through is `findOutputFiles` which is really only calling `evaluateFiles` on all the output expressions. `evaluateFiles` is a method on `WomExpression` and already being provided by WDL through `WdlWomExpression` so I *think* we might not really need this ? :/

danbills:
going to use approach suggested by @Horneth in https://github.com/broadinstitute/cromwell/pull/2656

PR #244, opened 2017-10-02
Expressions Checkpoint 

danbills:
This is a checkpoint on expressions intended to provide focus on a few changes:

These are:

 * retain typed representations of an ECMAScript function and expression all the way until evaluation.
 * introduce `File` constructor that omits `class` parameters
 * Change CwlExpressionCommandPart to take an expression instead of String.
 * Introduce `Expression` coproduct to represent ECMAScript function and expression.


cwl/src/main/scala/cwl/CwlFile.scala
```diff
@@ -71,12 +72,8 @@ case class CommandLineTool private(
...
 
   object ArgumentToId extends Poly1 {
-    implicit def ecmaScript = at[ECMAScript] {
-      _.value
-    }
-
-    implicit def ecmaFunction = at[ECMAFunction] {
-      _.value
+    implicit def ecmaScript:Case.Aux[Expression, String] = at[Expression] {
+      _ => ???
```
danbills, line 76: I think we came up with a good way to generate an ID for a `CommandLineTool => TaskDefinition`, but it seems it has not been implemented yet.  In any event I don't think an expression is a good candidate for an ID.  I'll create an issue if none has been filed.

cwl/src/main/scala/cwl/ExpressionEvaluator.scala
```diff
@@ -1,35 +1,53 @@
...
+  def evalExpression(expression: ECMAScriptExpression, parameterContext: ParameterContext): WdlValue = {
+    val ECMAScriptExpressionRegex = ECMAScriptExpressionWitness.value.r
+    expression.value match {
+      case ECMAScriptExpressionRegex(script) => JsUtil.eval(script, paramValues(parameterContext))
+      case _ => throw new RuntimeException(s"Expression was unable to be matched to Regex. This is never supposed to happen thanks to our JSON parsing library")
+    }
+  }
+
+  def evalFunction(function: ECMAScriptFunction, parameterContext: ParameterContext): WdlValue = {
+    val ECMAScriptFunctionRegex = ECMAScriptFunctionWitness.value.r
```
kshakir, line 32: Aha! I was wondering how to reuse a regex witness at compile time and runtime. Makes sense now. cc @mcovarr

kshakir (approved):

cwl/src/main/scala/cwl/CwlWomExpression.scala
```diff
@@ -24,42 +24,24 @@ sealed trait CwlWomExpression extends WomExpression {
...
-case class CommandOutputExpression(commandOutputParameter: CommandOutputParameter,
+case class CommandOutputExpression(outputBinding: CommandOutputBinding,
                                    override val cwlExpressionType: WdlType) extends CwlWomExpression {
   override def evaluateValue(inputValues: Map[String, WdlValue], ioFunctionSet: IoFunctionSet): ErrorOr[WdlValue] = {
     val parameterContext = ParameterContext.Empty.withInputs(inputValues, ioFunctionSet)
 
-    commandOutputParameter.outputBinding match {
-      case Some(outputBindingValue) =>
+    outputBinding match {
+      case outputBindingValue =>
```
Horneth, line 25: this match seems unnecessary now ?

cwl/src/main/scala/cwl/ParameterContext.scala
```diff
@@ -11,4 +11,7 @@ object ParameterContext {
   )
 }
 
-case class ParameterContext(inputs: WdlValue, self: WdlValue, runtime: WdlValue)
+case class ParameterContext(
+                             inputs: WdlValue = WdlOptionalValue(WdlNothingType, None),
```
Horneth: ToL: I hope we'll be able to narrow down the type a bit more than `WdlNothingType` later 😄 
`inputs`and `runtime` should be `WdlMap` I think ?

Horneth (approved):

cwl/src/main/scala/cwl/ParameterContext.scala
danbills: hm I think this code is unnecessary.

danbills: I.e. It's unnecessary given `Empty` and `withInputs`.

PR #250, opened 2017-10-05
[Pregull] WomIdentifier

Horneth:
Replaces `name: String` by an `identifier: WomIdentifier` that contains both a `localName` and a `fullyQualifiedName`. This is not strictly speaking necessary for WOM but useful in Cromwell to uniquely identify nodes and for reporting. 
Naming suggestions for `LocalName` and `FullyQualifiedName` welcome :D

It needs some clean up but the main idea is there. Seagull at will

### DIFF
--- a/cwl/src/main/scala/cwl/CwlExpressionCommandPart.scala
+++ b/cwl/src/main/scala/cwl/CwlExpressionCommandPart.scala
@@ -13,7 +13,7 @@ case class CwlExpressionCommandPart(expr: Expression) extends CommandPart {
                            functions: IoFunctionSet,
                            valueMapper: (WdlValue) => WdlValue): String = {
 
-    val pc = ParameterContext.Empty.withInputs(inputsMap.map({ case (key, value) => key.asString -> value }), functions)
+    val pc = ParameterContext.Empty.withInputs(inputsMap.map({ case (LocalName(localName), value) => localName -> value }), functions)
 
     val wdlValue: WdlValue = expr.fold(EvaluateExpression).apply(pc)
 
--- a/cwl/src/main/scala/cwl/CwlExpressionCommandPart.scala
+++ b/cwl/src/main/scala/cwl/CwlExpressionCommandPart.scala
@@ -5,14 +5,15 @@ import wom.CommandPart
 import wom.expression.IoFunctionSet
 import CwlWomExpression._
 import wdl4s.cwl.EvaluateExpression
+import wom.graph.LocalName
 
 
 case class CwlExpressionCommandPart(expr: Expression) extends CommandPart {
-  override def instantiate(inputsMap: Map[String, WdlValue],
-                            functions: IoFunctionSet,
-                            valueMapper: (WdlValue) => WdlValue): String = {
+  override def instantiate(inputsMap: Map[LocalName, WdlValue],
+                           functions: IoFunctionSet,
+                           valueMapper: (WdlValue) => WdlValue): String = {
 
-    val pc = ParameterContext.Empty.withInputs(inputsMap, functions)
+    val pc = ParameterContext.Empty.withInputs(inputsMap.map({ case (key, value) => key.asString -> value }), functions)
 
     val wdlValue: WdlValue = expr.fold(EvaluateExpression).apply(pc)
 
--- a/cwl/src/main/scala/cwl/WorkflowStep.scala
+++ b/cwl/src/main/scala/cwl/WorkflowStep.scala
@@ -61,7 +61,7 @@ case class WorkflowStep(
                      workflowInputs: Map[String, GraphNodeOutputPort]): Checked[Set[GraphNode]] = {
 
     // To avoid duplicating nodes, return immediately if we've already covered this node
-    val haveWeSeenThisStep: Boolean = knownNodes.collect { case TaskCallNode(name, _, _, _) => name }.contains(unqualifiedStepId)
+    val haveWeSeenThisStep: Boolean = knownNodes.collect { case TaskCallNode(name, _, _, _) => name.localName.asString }.contains(unqualifiedStepId)
 
     if (haveWeSeenThisStep) Right(knownNodes)
     else {
@@ -210,7 +210,7 @@ case class WorkflowStep(
       for {
         stepInputFold <- in.foldLeft(WorkflowStepInputFold.emptyRight)(foldStepInput)
         inputDefinitionFold <- taskDefinition.inputs.foldMap(foldInputDefinition(stepInputFold.stepInputMapping)).toEither
-        callAndNodes = callNodeBuilder.build(unqualifiedStepId, taskDefinition, inputDefinitionFold)
+        callAndNodes = callNodeBuilder.build(WomIdentifier(unqualifiedStepId), taskDefinition, inputDefinitionFold)
       } yield stepInputFold.generatedNodes ++ callAndNodes.nodes ++ knownNodes
     }
   }
@@ -246,7 +246,8 @@ object WorkflowStep {
   implicit class EnhancedWorkflowStepInput(val workflowStepInput: WorkflowStepInput) extends AnyVal {
     def toExpressionNode(sourceMappings: Map[String, OutputPort]): ErrorOr[ExpressionNode] = {
       val womExpression = PlaceholderWomExpression(sourceMappings.keySet, WdlAnyType)
-      ExpressionNode.linkWithInputs(workflowStepInput.id, womExpression, sourceMappings)
+      val identifier = WomIdentifier(workflowStepInput.id)
+      ExpressionNode.linkWithInputs(identifier, womExpression, sourceMappings)
     }
   }
 
--- a/cwl/src/main/scala/cwl/WorkflowStep.scala
+++ b/cwl/src/main/scala/cwl/WorkflowStep.scala
@@ -48,7 +48,7 @@ case class WorkflowStep(
 
   def fileName: Option[String] = run.select[String]
 
-  val unqualifiedStepId = Try(WorkflowStepId(id)).map(_.stepId).getOrElse(id)
+  val unqualifiedStepId = WomIdentifier(Try(WorkflowStepId(id)).map(_.stepId).getOrElse(id))
 
   /**
     * Generates all GraphNodes necessary to represent call nodes and input nodes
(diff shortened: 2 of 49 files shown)


wom/src/main/scala/wdl/Scope.scala
```diff
@@ -105,6 +106,18 @@ trait Scope {
   lazy val fullyQualifiedName = {
     (ancestry.filter(_.appearsInFqn).map(_.unqualifiedName).reverse :+ unqualifiedName).mkString(".")
   }
+  
+  lazy val womIdentifier = {
+    // Limit the fully qualified name to the parent workflow, if it exists.
+    // The reason for this is if this scope comes from an imported namespace,
+    // the name of the namespace will be part of the FQN. While useful to construct the hierarchy and dependencies,
+    // this is not desirable when being used by underlying engines.
```
kshakir: Use a block comment `/* … */` instead of multiple single line comments.

kshakir (approved):
I _think_ I follow what's going on here using FQN for some external nodes, and LN for others like expressions... so LGTM!

wom/src/main/scala/wdl/Scope.scala
mcovarr: But please do not use Scaladoc `/**` on locals as that produces fatal (!) "unmoored" errors during doc generation.

Horneth:
@kshakir yep that's exactly that. I found it unnecessarily painful to figure out a meaningful FQN for *every* node, so the external facing ones get a good one that we can use and the internal ones get the same as the "local" name they already had.

wom/src/main/scala/wdl/Scope.scala
Horneth: ok

cwl/src/main/scala/cwl/WorkflowStep.scala
```diff
@@ -61,7 +61,7 @@ case class WorkflowStep(
                      workflowInputs: Map[String, GraphNodeOutputPort]): Checked[Set[GraphNode]] = {
 
     // To avoid duplicating nodes, return immediately if we've already covered this node
-    val haveWeSeenThisStep: Boolean = knownNodes.collect { case TaskCallNode(name, _, _, _) => name }.contains(unqualifiedStepId)
+    val haveWeSeenThisStep: Boolean = knownNodes.collect { case TaskCallNode(name, _, _, _) => name.localName.asString }.contains(unqualifiedStepId)
```
mcovarr: `id` or `identifier` would probably be more appropriate variable names for a `WomIdentifier` than `name`

wom/src/main/scala/wom/graph/WomIdentifier.scala
```diff
@@ -0,0 +1,36 @@
+package wom.graph
+
+/**
+  * Identifies a node in its local context.
+  * Used in during graph construction to link nodes together and validate the graph.
+  */
+case class LocalName(private val value: String) {
+  def asString: String = value
+  def combineToLocalName(other: String) = LocalName(s"$value.$other")
```
mcovarr: are these `foo.bar` constructions not too WDL specific?

cwl/src/main/scala/cwl/CwlExpressionCommandPart.scala
```diff
@@ -5,14 +5,15 @@ import wom.CommandPart
...
 case class CwlExpressionCommandPart(expr: Expression) extends CommandPart {
-  override def instantiate(inputsMap: Map[String, WdlValue],
-                            functions: IoFunctionSet,
-                            valueMapper: (WdlValue) => WdlValue): String = {
+  override def instantiate(inputsMap: Map[LocalName, WdlValue],
+                           functions: IoFunctionSet,
+                           valueMapper: (WdlValue) => WdlValue): String = {
 
-    val pc = ParameterContext.Empty.withInputs(inputsMap, functions)
+    val pc = ParameterContext.Empty.withInputs(inputsMap.map({ case (key, value) => key.asString -> value }), functions)
```
mcovarr: ToL you could pattern match `LocalName(key)` if that field wasn't private (why is it private?).

wom/src/main/scala/wdl/Declaration.scala
```diff
@@ -120,21 +120,21 @@ object Declaration {
...
+        expressionNode <- ExpressionNode.linkWithInputs(decl.womIdentifier, womExpression, uninstantiatedExpression.inputMapping)
       } yield IntermediateValueDeclarationNode(expressionNode)
     }
 
     def workflowOutputAsGraphOutputNode(wdlExpression: WdlExpression) = {
       val womExpression = WdlWomExpression(wdlExpression, None)
       for {
         uninstantiatedExpression <- WdlWomExpression.findInputsforExpression(inputName, womExpression, localLookup, outerLookup)
-        graphOutputNode <- ExpressionBasedGraphOutputNode.linkWithInputs(inputName, womExpression, uninstantiatedExpression.inputMapping)
+        graphOutputNode <- ExpressionBasedGraphOutputNode.linkWithInputs(decl.womIdentifier, decl.wdlType, womExpression, uninstantiatedExpression.inputMapping)
```
mcovarr, line 131: just pass the `decl` rather than its disassembled parts?

wom/src/main/scala/wdl/Declaration.scala
```diff
@@ -120,21 +120,21 @@ object Declaration {
...
       } yield GraphOutputDeclarationNode(graphOutputNode)
     }
 
     decl match {
-      case Declaration(opt: WdlOptionalType, _, None, _, _) => Valid(InputDeclarationNode(OptionalGraphInputNode(inputName, opt, decl.fullyQualifiedName)))
-      case Declaration(_, _, None, _, _) => Valid(InputDeclarationNode(RequiredGraphInputNode(inputName, decl.wdlType, decl.fullyQualifiedName)))
+      case Declaration(opt: WdlOptionalType, _, None, _, _) => Valid(InputDeclarationNode(OptionalGraphInputNode(decl.womIdentifier, opt)))
+      case Declaration(_, _, None, _, _) => Valid(InputDeclarationNode(RequiredGraphInputNode(decl.womIdentifier, decl.wdlType)))
```
mcovarr, line 137: here too

wom/src/main/scala/wdl/Scope.scala
mcovarr: And the actual content of the comment is very nicely written BTW. 🙂

wom/src/main/scala/wom/graph/CallNode.scala
```diff
@@ -43,17 +43,26 @@ final case class WorkflowCallNode private(override val name: String,
 
 object TaskCall {
   def graphFromDefinition(taskDefinition: TaskDefinition): ErrorOr[Graph] = {
-    def linkOutput(call: GraphNode)(output: OutputDefinition): ErrorOr[GraphNode] = call.outputByName(output.name).map(out => PortBasedGraphOutputNode(output.name, output.womType, out))
+    val taskDefinitionLocalName = LocalName(taskDefinition.name)
+    
+    // Creates an identifier for an input or an output
+    // The localName is the name of the input or output
+    // The FQN combines the name of the task to the name of the input or output
+    def identifier(name: LocalName) = WomIdentifier(name, taskDefinitionLocalName.combineToFullyQualifiedName(name))
```
mcovarr, line 51: so this identifier isn't really *fully* qualified, it's just qualified to the local name of the task definition?

wom/src/main/scala/wom/graph/Graph.scala
```diff
@@ -58,6 +59,26 @@ object Graph {
...
+      .toList // Important since nodes is a Set, we don't want duplicates to disappear automatically when mapping to FQN
+      .map(_.identifier.fullyQualifiedName)
+      .groupBy(identity)
+      .collect({
+        case (fqn, list) if list.lengthCompare(1) > 0 => fqn
+      }).toList match {
+      case Nil => ().validNel
+      case head :: tail =>
+        NonEmptyList.of(head, tail: _*).map(fqn => s"Two or more nodes have the same FullyQualifiedName: ${fqn.asString}").invalid
+    }
```
mcovarr: holy smokes

wom/src/main/scala/wom/graph/GraphNode.scala
```diff
@@ -1,15 +1,19 @@
...
+import wom.graph.GraphNodePort.InputPort
 
 trait GraphNode {
+  def identifier: WomIdentifier
 
-  def name: String
+  /**
+    * Alias for identifier.localName.asString
+    */
+  final def name: String = identifier.localName.asString
```
mcovarr: maybe call it `localName`?

mcovarr (approved):

wom/src/main/scala/wom/graph/WomIdentifier.scala
Horneth: Yeah I thought of leaving this abstract and have each language define their `combine` but I convinced myself that the `.` separation could be the `WOM` separation too. Since Cromwell will use those FQNs for metadata etc it would be nice to have a consistent "look" between workflows for FQNs, even from different language.

cwl/src/main/scala/cwl/CwlExpressionCommandPart.scala
Horneth: Well I was trying to make it harder to get the String value so that we'd try to work with `LocalName` and `FullyQualifiedName` to get more type safety but yeah it seems a bit counter productive :p

wom/src/main/scala/wom/graph/CallNode.scala
Horneth: yes, although this method creates a CallNode and a Graph purely from a TaskDefinition. So the name of this task definition + name of the output is as much as we're going to get here which could be another name for "fully qualified" 😄

cwl/src/main/scala/cwl/CwlExpressionCommandPart.scala
geoffjentry: this line of commenting reminded me i keep meaning to say that i'd love the WOM effort to be the impetus for properly strongly typing LQNs and FQNs everywhere ...

Horneth: Yeah that's an effort towards that

wom/src/main/scala/wdl/Declaration.scala
Horneth: Actually no :p 
`ExpressionBasedGraphOutputNode` is a wom object and it can't know about WDL `Declaration` so we need to give it the identifier and type separately

PR #252, opened 2017-10-05
HttpImport Resolver (Phase 1 - sync)

kcibul:
With the fast moving pace of wdl4s & cromwell, I wanted to split this work into two phases:
 - P1: implement basic http import functionality in a sync manner, since that's the current interface (smaller change)
 - P2: move to async, which requires making much of WdlNamespace async and all it's clients cope with that (much bigger change)

P1 will let users (Geraldine) make sure this works for her use case and get some road time without waiting for the P2 (which just makes it more scalable)

I used the sttp library for http access because it will let us move between Try/Future/IO pretty easily depending on whatever the prevailing wisdom is during P2.

I chose to use a mock http server for testing because it let me assert the correct functionality in a lightweight way with specific failure modes instead of writing many centaur tests to cover all those same cases.  GET on HTTP is stable, so it's different than mocking out a complex http API with behavior

* non-async version of functioning code (async is phase II)
* tests using local mock http server (setup/teardown in test)
* logback-test.xml created to silence noisy netty debug log messages in tests


wom/src/test/scala/wdl4s/wdl/WdlWorkflowHttpImportSpec.scala
```diff
@@ -0,0 +1,108 @@
...
+
+
+  "The httpResolver" should "not resolve an invalid URL " in {
+    val wf = tinyWorkflow("ht://foobar")
+    val ns = WdlNamespaceWithWorkflow.load(wf, httpResolver)
+    ns.isFailure shouldBe true
+  }
+
+  it should "resolve an http URL" in {
+    val wf = tinyWorkflow( "http://localhost:" + mockServer.getPort + "/hello.wdl")
```
Horneth: looks like it would be worth having a `host` variable for this `"http://localhost:" + mockServer.getPort` since it's repeated a few times

wom/src/test/resources/logback-test.xml
```diff
@@ -0,0 +1,18 @@
...
+    </appender>
+
+    <logger name="io.netty" level="INFO"/>
+    <logger name="org.asynchttpclient.netty" level="INFO"/>
+    <logger name="org.mockserver" level="WARN"/>
+
+    <root level="debug">
+        <appender-ref ref="STDOUT" />
+    </root>
+</configuration>
```
mcovarr: \n

wom/src/main/scala/wdl/WdlNamespace.scala
```diff
@@ -563,6 +571,23 @@ object WdlNamespace {
...
   }
+
+  def httpResolver(str: String): WorkflowSource =
+    httpResolverWithHeaders(Map.empty[String,String])(str)
+
+  def httpResolverWithHeaders(headers: Map[String,String])(str: String): WorkflowSource = {
+    implicit val sttpBackend = AsyncHttpClientCatsBackend[IO]()
+
+    val responseIO : IO[Response[String]] =
+      sttp.get(uri"$str").headers(headers)send()
```
mcovarr: I don't get this syntax...

wom/src/main/scala/wdl/WdlNamespace.scala
```diff
@@ -563,6 +571,23 @@ object WdlNamespace {
       throw new IllegalArgumentException(s"$str is not a valid import")
     }
   }
+
+  def httpResolver(str: String): WorkflowSource =
+    httpResolverWithHeaders(Map.empty[String,String])(str)
+
+  def httpResolverWithHeaders(headers: Map[String,String])(str: String): WorkflowSource = {
```
mcovarr: some comments would be helpful

wom/src/test/scala/wdl4s/wdl/WdlWorkflowHttpImportSpec.scala
```diff
@@ -0,0 +1,108 @@
...
+    ).respond(
+      response().withStatusCode(301).withHeader("Location","/hello.wdl"))
+
+    mockServer.when(
+      request().withPath("/none.wdl")
+    ).respond(
+      response().withStatusCode(404)
+    )
+
+    () // explicitly return unit
```
mcovarr, line 68: This might have been a(n unpleasant) surprise, but it's all over the code base and probably doesn't warrant a comment. 🙂

wom/src/test/resources/logback-test.xml
```diff
@@ -0,0 +1,18 @@
+<configuration>
```
mcovarr, line 1: Did you mean to check this file in or was this just for testing?

mcovarr (approved):
Very cool!

wom/src/test/scala/wdl4s/wdl/WdlWorkflowHttpImportSpec.scala
geoffjentry: Yeah, our linters are turned up to 11 :)

wom/src/test/resources/logback-test.xml
mcovarr: oh nm, you address this in your description. I'm not used to such well-written descriptions, I wasn't expecting that. 😛

Horneth (approved):
nice !

wom/src/main/scala/wdl/WdlNamespace.scala
kcibul: huh... me neither, looks like the "." went missing but it somehow still works.  I'll put the . back

PR #255, opened 2017-10-11
hotfix version of mainline functionality for http imports

kcibul:
... also had to fix a deprecation error due to newer cats being pulled in


danbills (approved):
LGTM, thanks for async-ify comment as that's my only concern.

wom/src/test/resources/logback-test.xml
```diff
@@ -0,0 +1,18 @@
...
+    </appender>
+
+    <logger name="io.netty" level="INFO"/>
+    <logger name="org.asynchttpclient.netty" level="INFO"/>
+    <logger name="org.mockserver" level="WARN"/>
+
+    <root level="debug">
+        <appender-ref ref="STDOUT" />
+    </root>
+</configuration>
```
cjllanwarne: Newline please!

cjllanwarne (approved):
Looks good to me!

wom/src/test/scala/wdl4s/wdl/WdlWorkflowHttpImportSpec.scala
```diff
@@ -0,0 +1,109 @@
...
+    ns.isFailure shouldBe true
+  }
+
+  it should "follow a redirect" in {
+    val wf = tinyWorkflow( s"$host/redirect.wdl")
+    val ns = WdlNamespaceWithWorkflow.load(wf, httpResolver)
+    ns.isFailure shouldBe false
+  }
+
+  it should "be able to supply a bearer token to a protected resource" in {
```
cjllanwarne: Is it worth also adding "not supply a bearer token to an inappropriate URL" test?

eg we don't want to be spamming non-https endpoints with our super secret tokens!

cjllanwarne: Sorry if this comment is a dupe (github things?)

Is it worth a `it should "not supply a bearer token to an unprotected resource"` test to make sure we wouldn't accidentally spam out tokens to non-https endpoints?

cjllanwarne:
Sorry if this comment is a dupe (github things?)

Is it worth a `it should "not supply a bearer token to an unprotected resource"` test to make sure we wouldn't accidentally spam out tokens to non-https endpoints just because we have some available in case we need them?

cjllanwarne:
In the same way, if we have two tokens for different sources, only the correct token should be sent to each source.

kcibul:
@cjllanwarne -- thanks for the comments.  I think those tests might be more appropriate for centaur because at this level it's just "can we supply a header" but it's up to the caller to decide what and when.  In the initial cromwell PR this will only be public URLs and no headers will be sent, but I was just thinking ahead because accessing protected URLs is going to be asked for soon

PR #256, opened 2017-10-31
Improving type evaluation and lookup

orodeh:
I ran into some issues while using wdl4s inside the dxWDL compiler. I thought I would contribute upstream the fixes I came up with. Note that these fixes may be inaccurate. They pass the existing  tests, and the new test that I added. 

I ran the existing tests like this:
```testOnly wdl.*```

The test `wdl.wom.WdlSubworkflowWomSpec` was converting a String to an Array[String] automatically. Is this correct? I modified it, and added braces.

I hope this helps.

### DIFF
--- a/wom/src/main/scala/wdl/WdlNamespace.scala
+++ b/wom/src/main/scala/wdl/WdlNamespace.scala
@@ -455,13 +455,18 @@ object WdlNamespace {
     invalidVariableReferences ++ typeMismatches
   }
 
-  private def lookupType(from: Scope)(n: String): WdlType = {
+  def lookupType(from: Scope)(n: String): WdlType = {
     val resolved = from.resolveVariable(n)
     resolved match {
       case Some(d: DeclarationInterface) => d.relativeWdlType(from)
       case Some(c: WdlCall) => WdlCallOutputsObjectType(c)
       case Some(s: Scatter) => s.collection.evaluateType(lookupType(s), new WdlStandardLibraryFunctionsType, Option(from)) match {
-        case Success(a: WdlArrayType) => a.memberType
+        case Success(a: WdlArrayType) =>
+          // Collection is an array
+          a.memberType
+        case Success(WdlMapType(kType,vType)) =>
+          // Collection is a map
+          WdlPairType(kType, vType)
         case _ => throw new VariableLookupException(s"Variable $n references a scatter block ${s.fullyQualifiedName}, but the collection does not evaluate to an array")
       }
       case Some(_: WdlNamespace) => WdlNamespaceType
--- a/wom/src/main/scala/wdl/expression/TypeEvaluator.scala
+++ b/wom/src/main/scala/wdl/expression/TypeEvaluator.scala
@@ -83,10 +83,13 @@ case class TypeEvaluator(override val lookup: String => WdlType, override val fu
           evaluate(a.getAttribute("lhs")).flatMap {
             case o: WdlCallOutputsObjectType =>
               o.call.outputs.find(_.unqualifiedName == rhs.getSourceString) match {
-                case Some(taskOutput) => 
-                  from map { source =>
-                    evaluate(taskOutput.requiredExpression.ast) map { t => DeclarationInterface.relativeWdlType(source, taskOutput, t) }
-                  } getOrElse evaluate(taskOutput.requiredExpression.ast)
+                case Some(taskOutput) =>
+                  val t = taskOutput.wdlType
+                  val relative = from match {
+                    case None => t
+                    case Some(scope) => DeclarationInterface.relativeWdlType(scope, taskOutput, t)
+                  }
+                  Success(relative)
                 case None => Failure(new WdlExpressionException(s"Could not find key ${rhs.getSourceString}"))
               }
             case WdlPairType(leftType, rightType) =>
@@ -130,4 +133,3 @@ case class TypeEvaluator(override val lookup: String => WdlType, override val fu
     }
   }
 }
-
--- a/wom/src/test/scala/wdl/DNAxTypeEvalTest.scala
+++ b/wom/src/test/scala/wdl/DNAxTypeEvalTest.scala
@@ -1,0 +1,182 @@
+package wdl
+
+import org.scalatest.{FlatSpec}
+import scala.util.{Failure, Success}
+import wdl.expression._
+import wdl.types._
+
+class DNAxTypeEvalTest extends FlatSpec {
+
+    val wdlCode =
+        """|task Add {
+           |  Int a
+           |  Int b
+           |
+           |  command {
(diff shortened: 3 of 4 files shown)

### CODE REVIEW

**Comment by geoffjentry** [comment]
Hi @orodeh -

First, this is great that you did this, thanks!

Second, now is the part where I'm a pain in the butt ;) We recently moved things from wdl4s directly into the Cromwell repo for developer sanity reasons.  In particular you'll want to look at https://github.com/broadinstitute/cromwell/tree/develop/wom/src and https://github.com/broadinstitute/cromwell/tree/develop/wdl/src

Tagging @mcovarr and @cjllanwarne who can help get you on the right path as necessary

File: wom/src/main/scala/wdl/WdlNamespace.scala [hunk 1]
```diff
@@ -455,13 +455,18 @@ object WdlNamespace {
...
     val resolved = from.resolveVariable(n)
     resolved match {
       case Some(d: DeclarationInterface) => d.relativeWdlType(from)
       case Some(c: WdlCall) => WdlCallOutputsObjectType(c)
       case Some(s: Scatter) => s.collection.evaluateType(lookupType(s), new WdlStandardLibraryFunctionsType, Option(from)) match {
-        case Success(a: WdlArrayType) => a.memberType
+        case Success(a: WdlArrayType) =>
+          // Collection is an array
+          a.memberType
+        case Success(WdlMapType(kType,vType)) =>
```
cjllanwarne, line 467: This could be covered by the custom unapply of `WdlArrayType`, eg this one-liner should work:
```
case Success(WdlArrayType(arrayType)) => arrayType
```

wom/src/test/scala/wdl/wom/WdlSubworkflowWomSpec.scala
```diff
@@ -18,7 +18,7 @@ class WdlSubworkflowWomSpec extends FlatSpec with Matchers {
         |  Int x
         |  call import_me.inner as inner { input: i = x }
         |  output {
-        |    Array[String] out = inner.out
+        |    Array[String] out = [inner.out]
```
cjllanwarne, line 21: It's awesome that these can now be statically type-checked!

TBH I think I'd prefer just fixing the type to `String` rather than adding the `[ ]`s, but if that means a lot of expectation fixing then I'm not too fussed

wom/src/main/scala/wdl/expression/TypeEvaluator.scala
```diff
@@ -83,10 +83,13 @@ case class TypeEvaluator(override val lookup: String => WdlType, override val fu
           evaluate(a.getAttribute("lhs")).flatMap {
             case o: WdlCallOutputsObjectType =>
               o.call.outputs.find(_.unqualifiedName == rhs.getSourceString) match {
-                case Some(taskOutput) => 
-                  from map { source =>
-                    evaluate(taskOutput.requiredExpression.ast) map { t => DeclarationInterface.relativeWdlType(source, taskOutput, t) }
-                  } getOrElse evaluate(taskOutput.requiredExpression.ast)
+                case Some(taskOutput) =>
+                  val t = taskOutput.wdlType
```
cjllanwarne, line 87: So this is using the declared output type rather than the type of the output expression? I like that - the output should be evaluated and then coerced into the declared type so IMO this is more accurate!

cjllanwarne (approved):
This looks correct to me, thanks!

Small comments but otherwise I expect this should map pretty directly into the new single repo version.

wom/src/main/scala/wdl/WdlNamespace.scala
mcovarr: ToL (i.e. unrelated to this PR) I would not have expected that. Perhaps we should have some "array-like" trait that the array and map types can share that would make this less surprising.

wom/src/main/scala/wdl/expression/TypeEvaluator.scala
mcovarr: Isn't the caller perhaps trying to ascertain if the expression actually can be coerced to the declared type?

wom/src/test/scala/wdl/DNAxTypeEvalTest.scala
```diff
@@ -0,0 +1,182 @@
...
+
+
+    // Figure out the type of an expression
+    def evalType(expr: WdlExpression, parent: Scope) : WdlType = {
+        expr.evaluateType(WdlNamespace.lookupType(parent),
+                          new WdlStandardLibraryFunctionsType,
+                          Some(parent)) match {
+            case Success(wdlType) => wdlType
+            case Failure(f) =>
+                System.err.println(s"could not evaluate type of expression ${expr.toWdlString}")
```
mcovarr, line 51: no printlns please 🙂

wom/src/test/scala/wdl/DNAxTypeEvalTest.scala
```diff
@@ -0,0 +1,182 @@
...
+                throw f
```
mcovarr, line 52: I'd generally prefer `fail`ing to `throw`ing throughout

wom/src/test/scala/wdl/DNAxTypeEvalTest.scala
```diff
@@ -0,0 +1,182 @@
...
+        val wf = ns.workflow
+
+        val copy2call:WdlCall = wf.findCallByName("Copy2") match {
+            case None => throw new Exception(s"Call Add not found in WDL file")
+            case Some(call) => call
+        }
+
+        val e1 = copy2call.inputMappings("src")
+        val t1 = evalType(e1, copy2call)
+        //System.err.println(s"type(${e1.toWdlString}) = ${t1.toWdlString}")
```
mcovarr, line 112: All printlns should go, especially commented out ones. 🙂

wom/src/test/scala/wdl/DNAxTypeEvalTest.scala
```diff
@@ -0,0 +1,182 @@
...
+        assert(t1 == WdlFileType)
```
mcovarr, line 113: `a should equal(b)` would print a more helpful message on failure than `assert(a == b)`, throughout

wom/src/test/scala/wdl/wom/WdlSubworkflowWomSpec.scala
mcovarr: The magic conversion of a single `Thing` to an `Array[Thing]` was a feature someone explicitly asked for way back when. I never liked this feature but I'm sure there's WDL out there that expects this to work.

orodeh:
I ported this pull request to the Cromwell tree,
https://github.com/broadinstitute/cromwell/pull/2807.

PR #257, opened 2017-11-12
fix subworkflow declarations

mcovarr:
Do not evaluate declarations in subworkflows too eagerly. Subworkflows are not tasks.

### DIFF
--- a/wom/src/main/scala/wdl4s/wdl/WdlCall.scala
+++ b/wom/src/main/scala/wdl4s/wdl/WdlCall.scala
@@ -1,8 +1,9 @@
 package wdl4s.wdl
 
+import lenthall.util.TryUtil
 import wdl4s.parser.WdlParser.{Ast, SyntaxError, Terminal}
 import wdl4s.wdl.AstTools.EnhancedAstNode
-import wdl4s.wdl.exception.{ValidationException, VariableLookupException, VariableNotFoundException}
+import wdl4s.wdl.exception.{VariableLookupException, VariableNotFoundException}
 import wdl4s.wdl.expression.WdlFunctions
 import wdl4s.wdl.types.WdlOptionalType
 import wdl4s.wdl.values.{WdlOptionalValue, WdlValue}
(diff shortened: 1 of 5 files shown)


wom/src/main/scala/wdl4s/wdl/examples/ex7.scala
```diff
@@ -36,7 +36,7 @@ object ex7 {
 
     ns.taskCalls.find( _.unqualifiedName == "a") foreach { call =>
       val wdlFunctions: CustomFunctions = new CustomFunctions
-      val evaluatedInputs = call.evaluateTaskInputs(inputs, wdlFunctions).get
+      val evaluatedInputs = call.evaluateCallDeclarations(inputs, wdlFunctions).get
```
kshakir, line 39: Each file in this folder is a compiled check of a README example. Thus any updates here should have an accompanying change there.

In this case, the corresponding block is [here](https://github.com/broadinstitute/wdl4s/blob/814d20317b835e3bb785e57d9c17cac2676a1dbb/README.md#user-content-instantiating-commands).

kshakir (approved):
One updated to the README needed, then good to go.

wom/src/main/scala/wdl4s/wdl/WdlCall.scala
```diff
@@ -231,11 +225,39 @@ sealed abstract class WdlCall(val alias: Option[String],
...
+        // Workflow declarations that are initialized can also be overridden by inputs in the `input:` block of a
+        // subworkflow call. These overrides must be identified so the overriding expressions can be evaluated.
+        val overriddenDeclarations = declarations filter {
+          d => d.expression.isDefined && suppliedInputNames.contains(d.unqualifiedName)
+        }
+        Success((declarationsRequiringInputs ++ overriddenDeclarations).toSeq)
+    }
+  }
+
 }
```
cjllanwarne: NYF: newline

cjllanwarne (approved):
Looks good to me, but I'd like to double check all the cases I mention in the centaur PR

